PR #98, opened 2015-08-17
igbmc-gempicker improved protocol

azazellochg:
* Added extra parameters for picking (GUI, Advanced level)
* References can be also a set of particles (e.g., reprojections), so pointerClass for references was modified
* Added help for the protocol
* Simple threshold check to stay in 0-1 range
* GPU-related options
* Other minor corrections of variable names and description

PS. I also propose to change module name to gempicker instead of igbmc (pyworkflow/em/packages/igbmc -> pyworkflow/em/packages/gempicker)

### DIFF
--- a/pyworkflow/em/packages/igbmc/protocol_gempicker.py
+++ b/pyworkflow/em/packages/igbmc/protocol_gempicker.py
@@ -46,17 +46,14 @@ class ProtGemPicker(em.ProtParticlePicking):
     def _defineParams(self, form):
         
         em.ProtParticlePicking._defineParams(self, form)
-        
         form.addParam('inputReferences', params.PointerParam, 
-                      pointerClass='SetOfAverages',
+                      pointerClass='SetOfParticles',
                       label='Input References', important=True,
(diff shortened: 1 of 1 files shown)


azazellochg:
I have realized that later. Indeed, the averages are also particles. In (https://github.com/azazellochg/scipion/commit/a4953703efab650c9e8c413d862686ca451a745a) I however specified both in case one wants to use reprojections from the 3D model - these are not averages

delarosatrevin:
Having both SetOfParticles and SetOfAverages is the same as having only SetOfParticles. When an input accepts SetOfParticles, it will show all its subclasses as valid input. Indeed re-projections of a 3D model are not averages, but neither particles. In this case there are conceptually more close to averages than to particles. In practice we want to avoid the user to show ALL the particles in a project when he/she is selecting the input for the template picking and it probably will not work with a full set of particles. So this is the main reason for the distinction between averages (or 3D re-projections) and particles for some input parameters. We think that having all possible particle sets as input can lead users to errors.

PR #133, opened 2015-11-23
Fix common misspellings in source code

azazellochg:
I used [codespell](https://github.com/lucasdemarchi/codespell) check and verified all changes manually, so that var names etc are not touched.


delarosatrevin:
Thanks for this check. Can you comment a bit more about how to use codespell? (just for the future or for use in the wiki)

azazellochg:
@delarosatrevin , ```codespell -h``` is pretty much self-explanatory. I have used ```codespell -w -i 3 DIRNAME``` to interactively choose the right correction and write into the files.

PR #149, opened 2015-12-04
ctffind3 output is always *_ctffind3.log file

azazellochg:


delarosatrevin:
Is it always in this way? I guess .out extension is there for some reason.
@josuegbl , can you check and comment about this?

azazellochg:
@delarosatrevin actually, ctffind3 itself does not produce any log files and prints everything to stdout. Relion, however, always create *_ctffind3.log files - and this cannot be changed by user.

For ctffind4 there are both .log and .txt available after Relion run. If ctffind4 is run independently, it produces only .txt files.

josuegbl:
@delarosatrevin and @azazellochg: out extension comes inherited from xmipp-2.4. @azazellochg has right when says "ctffind3 itself does not produce any log files and prints everything to stdout". Since relion create *_ctffind3.log files, my suggestion is include .log but not delete .out.

delarosatrevin:
We have merge the Pull Request, but we kept the '.out' to maintain xmipp2.4 compatibility in ctffind outputs.
See 001e1697 commit.

PR #160, opened 2015-12-09
first version of Gctf program implementation

azazellochg:
Only simple CTF estimation from mics is implemented (no particles/movies options yet).
I have tested so far CTF import from gctf, estimation/reestimation, also there is a test protocol.

PS. the code is mostly copied from ctffind protocol, so feel free to correct authors info :)


delarosatrevin:
Great! I will update the tgz to our server.

PR #168, opened 2015-12-10
fix ctffind viewer

azazellochg:
Otherwise, the last "2sigma" graph is not shown


delarosatrevin:
@josuegbl , Can you check about this?

PR #200, opened 2016-01-11
removed deprecated data.py files from some packages

josuegbl:


delarosatrevin:
I restored this branch because @josuegbl forgot to remove some imports (and I didn't properly check it)

PR #207, opened 2016-01-12
Fix cl2d viewer

delarosatrevin:


delarosatrevin:
@cossorzano , I have updated the PR branch with latest changes from v1.0, please check if now the cl2d viewer is working.

PR #210, opened 2016-01-13
fix error in gctf plot, introduced in 17c0e8f

azazellochg:
fix error in gctf plot, introduced in 17c0e8f


delarosatrevin:
PR have been merge, you can freely delete this branch.

PR #214, opened 2016-01-13
fixed undefined [one] var in spider pairwise align protocol

### LINKED ISSUE #127: center1.spi is missing for spider align pairwise protocol
When using center of gravity = RT180, the spider protocol fails at [this](https://github.com/biocompwebs/scipion/blob/master/pyworkflow/em/packages/spider/scripts/mda/pairwise.msa#L475) point with an error:

>00066:    *** ERROR: PROCEDURE FILE.spi DOES NOT EXIST: /home/sharov/soft/scipion/software/em/spider/spider/proc/center1.spi

The required spider script (not adapted) can be found [here](http://spider.wadsworth.org/spider_doc/spider/docs/techs/ranconical/center1.rct)

azazellochg:
this closes #127

### DIFF
--- a/pyworkflow/em/packages/spider/scripts/mda/pairwise.msa
+++ b/pyworkflow/em/packages/spider/scripts/mda/pairwise.msa
@@ -470,9 +470,8 @@ if([cg-option].eq.1) then
 elseif([cg-option].eq.2) then
     vm
     echo "Centering average by rotating 180 degrees and aligning"
-    
     ; center by rotating by 180 degrees and self-alignment
-    @center1(-xshift-cg],[yshift-cg])
+    @center1([xshift-cg],[yshift-cg])
     [depth_dir]{***[next-depth]}/[intermed_avg][one]
     [pair_dir]/[centered_avg][one]
 else
--- a/pyworkflow/em/packages/spider/scripts/mda/center1.msa
+++ b/pyworkflow/em/packages/spider/scripts/mda/center1.msa
@@ -1,5 +1,7 @@
 (x21,x22)
 
+[one] = 1 ; dummy variable
+
 fr
 [input]?file to be centered?
 fr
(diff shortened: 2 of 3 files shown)


pyworkflow/em/packages/spider/scripts/mda/pairwise.msa
```diff
@@ -470,9 +470,8 @@ if([cg-option].eq.1) then
 elseif([cg-option].eq.2) then
     vm
     echo "Centering average by rotating 180 degrees and aligning"
-    
     ; center by rotating by 180 degrees and self-alignment
-    @center1(-xshift-cg],[yshift-cg])
```
delarosatrevin: Have you notice the minus sign here? @taputechnic, something to say?

delarosatrevin:
Thanks @azazellochg  for this fix, I have commented about a minus sign when calling the center procedure. After this we can merge...maybe I will add a test for this case.

pyworkflow/em/packages/spider/scripts/mda/pairwise.msa
taputechnic: That minus sign should be an open square bracket.  I hadn't checked it 
with the RT180 option.

I'm unable to get Scipion to "see" the changes I've made to pairwise.msa 
on the computer I'm on, so I can't verify that this is the only problem.

-Tapu

azazellochg:
@delarosatrevin , I'm almost certain it was a typo, since:
1) in original script http://spider.wadsworth.org/spider_doc/spider/docs/techs/ranconical/pairwise.rct it is absent
2) all vars passed to procedure in spider must be in []
3) this is not a correct arithmethic expression and -xshift-cg variable is not defined

pyworkflow/em/packages/spider/scripts/mda/pairwise.msa
delarosatrevin: @taputechnic , I think we are out of time to update the spider scripts before the v1.0.0 release, maybe for the next release. I would help you anyway to get scipion updated. I have added a test for alignment and the RT180 option seems to be working.

delarosatrevin:
Thanks again @azazellochg, RT180 option seems to be working! And we have now a dedicated test for spider alignment protocols.

PR #231, opened 2016-01-15
Fix some issues with ResMap resolution slices figure

josuegbl:
added four resmap tests.


delarosatrevin:
@josuegbl , I have finish with the changes/fixes...double check everything is working as expected.
Before merge, we need to create a tgz with new ResMap and update the install script.

PR #235, opened 2016-01-18
use cpu when gpu is selected

azazellochg:
use cpu when gpu is selected


pyworkflow/em/packages/igbmc/protocol_gempicker.py
```diff
@@ -205,7 +205,7 @@ def runGemPickerStep(self, mode, args):
         args += ' --mode=%d' % mode
         
         if self.useGPU:
```
delarosatrevin: @azazellochg , Is the number of threads also used with GPU? If it is the case, then we could re-write the code like:
nGPUs = self.numberOfGPUs.get() if self.useGPU else 0
nThreads = self.numberOfThreads.get()

Could you please also add yourself as author in the protocol header?

azazellochg: @delarosatrevin , yes. Thanks for the correction!

PR #238, opened 2016-01-19
fixing protocol structure map error

mohsenkazemi:
The error relating to Sklearn library is temporarily fixed by replacing "import sklearn" in the body of the protocol and validate whether this library is installed before running the protocol. Of course, I have to add a script for installing this library in the error message after discussing with Jesus.


delarosatrevin:
@mohsenkazemi , this should be merged into devel, not master.
Please reopen the Pull Request.

PR #239, opened 2016-01-19
temporarily fixing structure mapping protocol

mohsenkazemi:
still I need to discuss with Jesus to add necessary scripts to install library in error message.


pyworkflow/em/packages/xmipp3/__init__.py
```diff
@@ -52,7 +52,7 @@
 from viewer_normalize_strain import XmippNormalizeStrainViewer
 from viewer_resolution3d import XmippResolution3DViewer
 from viewer_split_volume import XmippViewerSplitVolume
-#from viewer_structure_mapping import XmippProtStructureMappingViewer
+from viewer_structure_mapping import XmippProtStructureMappingViewer
```
delarosatrevin: @mohsenkazemi , XmippProtStructureMappingViewer is imported twice

delarosatrevin:
I think (also discussed with @cossorzano ) that the protocol (and viewer) should be moved to the nma submodule under xmipp3.

The results files also needs some review (ask @cossorzano )

Finally, a bit of coding style can be improved (ask @delarosatrevin)

PR #288, opened 2016-02-01
Update Devel with lastest v1.0.0 Release fixes

delarosatrevin:


delarosatrevin:
I need to update from a different branch.

PR #303, opened 2016-02-09
Co multiple fscs

cossorzano:


delarosatrevin:
@cossorzano , before merging the branch into **devel** I would suggest:

1.  Add at least add one test for the protocol 
2. Synchronize with **fsc** branch from @rmarabini where he has added the FSC object and viewer, just to centralize the plotting code.

What do you think?

delarosatrevin:
The help of the protocols is wrong and the authors of the viewer.
There are several places that need to follow our code style guide. I don't mind to change that as long as I have a test that I can automatically run.
Again, since this branch is very related with @rmarabini **fsc**, we could merge both and do a single merge into **devel**

delarosatrevin:
@cossorzano , I have done the dirty work for you...I created the test and did a bit of refactoring of the protocol. The only different is that volumes are converted always, but the code is clearer. We could delete the volumes at the end if space is a concern. You should are more validation to the test.

PR #304, opened 2016-02-09
Devel

rmarabini:
FSC isn object.

  data.py -> create fsc class
  relion define, protocol_resolution3D -> create an fsc object at the end 
  viewer.py -> create fsc visualizer. It will send all the FSC curves to the same windoe


delarosatrevin:
@rmarabini , the pull request should use as base the branch **devel**.
This means that you want to merge **fsc** into **devel**. 
For the other way around is just a simply merge that you can do as usual
(after pull in **devel**).
So I will close this Pull Request....what do you think about merging your branch
with the @cossorzano  #303 ?

PR #306, opened 2016-02-11
Fix gctf citation and update version

azazellochg:


delarosatrevin:
@azazellochg, Thanks for the update.

PR #308, opened 2016-02-12
first version of Gautomatch protocol

azazellochg:


delarosatrevin:
@azazellochg , I have copy the Gautomatch binary to the required folder, could you test if the following works?
scipion install Gautomach_v0.50 
scipion install --help # should list Gautomach as well

Moreover, could you add some tests for the protocol? I would be better to re-use some of the existing datasets (such as the one of gEMpicker or another with micrographs and templates)

delarosatrevin:
In the bibtex, Zhang2016 is both used for gctf and gautomatch, I think that does not affect now since are in different packages, but maybe would be better to assign different keys.

azazellochg:
@delarosatrevin , 

Building Gautomatch_v0.50 ...
wget -nv -c -O software/em/Gautomatch_v0.50.tgz.part http://scipion.cnb.csic.es/downloads/scipion/software/em/Gautomatch_v0.50.tgz
http://scipion.cnb.csic.es/downloads/scipion/software/em/Gautomatch_v0.50.tgz:
2016-02-12 17:47:31 ERROR 404: Not Found.

I will add the tests soon.

delarosatrevin:
Now the installation should works. Just renamed .tar.gz to .tgz

azazellochg:
Yep, now it is works

delarosatrevin:
By the way, I also added the Gctf_v0.50.tgz file, I removed the examples (reduced from 125Mb to ~7Mb)

azazellochg:
@delarosatrevin ,thanks. But you should remove bin folder and move binaries into Gctf_v0.50, otherwise the protocol won't find them. Of we can add 'bin' to the protocol.

delarosatrevin:
Maybe we should add the 'bin' to the protocol, just to keep the files structure similar to the original (without the 'examples' I know). What do you think?

azazellochg:
@delarosatrevin , agreed. done

delarosatrevin:
Thanks!

azazellochg:
Please do not merge yet, there is a bug with template conversion.

azazellochg:
Now it seems to be ready.

delarosatrevin:
Finally I have been able to launch the Gautomatch test. I was surprised that it only picked side views even if the references are both side and top views. Do you know why?

Moreover, I would like to add the wizard in this same branch. I think it would be very helpful to tune the parameters.

azazellochg:
@delarosatrevin , I also noticed that. I will have a look at the test again with last changes I made.

azazellochg:
@delarosatrevin , I am starting to understand how it works :)
I have fixed the tests, now it picks everything properly.

delarosatrevin:
Great! I will re-run the test.

delarosatrevin:
There are still some prints that looks like debugging...could we remove them?

azazellochg:
@delarosatrevin , I found just one - removed.

PR #311, opened 2016-02-13
Fix protocols description, removed unused docstrings

azazellochg:
Corrected / improved desc of most protocols. Exception is xmipp3 - multireference_aligneability. It has wrong description and I have no idea what it does..

### DIFF
--- a/pyworkflow/em/packages/bsoft/bsoft.py
+++ b/pyworkflow/em/packages/bsoft/bsoft.py
@@ -23,13 +23,10 @@
 # *  e-mail address 'azaldivar@cnb.csic.es'
 # *
 # **************************************************************************
-"""
-This sub-package will contains Eman3.0 specific protocols
-"""
 
 import os
 
-from pyworkflow.utils import environAdd
+from pyworkflow.utils import environAdd, Environ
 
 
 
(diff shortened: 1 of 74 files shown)


delarosatrevin:
Thanks a lot! I have done some minor corrections:
1) Particle picking is manual/supervised, automatic is a separated protocol
2) Tilt-pair picking is only manual
4) Removed some doc strings at after the license, we started using that, but at the end are not useful and usually out-of-date due to copy-paste
3) Removed some blank lines, added some spaces

Double check my changes and tell me if are OK for you.

azazellochg:
Looks OK! Should we remove doc lines after the license everywhere? (This sub-package contains ...), since every class has a description?

delarosatrevin:
You are right, since the class have a description those doc lines are not longer useful for protocols. 
I was trying to avoid cleaning that for all protocols...but it is done now. I removed
"This sub-package contains..."
"This sub-package will contains..."
Only left the ones in some __init__.py files.

Could you double check? Thanks again for all your help.

azazellochg:
You are welcome. You missed just one file :) Now it looks clean, you can merge it.

delarosatrevin:
Great! I just submitted minor formatting in  xmipp3 - multireference_aligneability.
I will merge this and check tomorrow the help for this one together with break-symmetry, another one that I think has a wrong description.

PR #312, opened 2016-02-14
Fix non ascii chars

azazellochg:
fix non-ascii chars introduced in the previous merge


delarosatrevin:
I just realized that right now and fixed directly in v1.0, please update.

PR #317, opened 2016-02-15
Fix relion viewer

azazellochg:
Raise exception if iter. selection is not provided. Please double check my corrections to the code.


delarosatrevin:
@azazellochg , I have done some changes:
1) All the message said 'iterations', that was an easy one.
2) Raising the Exception was not very helpful for the user since was printed in the console
3) I stored the errors in a list...and then when any visualization option is clicked, it will show the error messages
4) I could not test for Frame selection
5) I also checked the empty string that is not the same as None

Thanks for pointing out this issue, good team work here!
Check if this works for you and let me know to merge the changes.

azazellochg:
@delarosatrevin , thank you for rewrting it in a more elegant way :+1: 
I have checked only on 2d/3d - have no frames to test, seems to work fine.

PR #325, opened 2016-02-18
Ctf viewer clean-up and PSD display fixed

delarosatrevin:
Main fixes:
- SetOfCtf visualization is the same if you double click the object or in the Analyze Results.
- CtfView is now centralized and the labels are passed if they are present in the data (xmipp labels only appears when xmipp results)
- The display CTF fitting work as well for import ctf from ctffind4
- The PSD profile now works with ctf:mrc trick.

@azazellochg , could you check if these changes work for you? I tested import but from scipion-ctffind4, could you test for original-ctffind4 and for relion-ctffind4?

### DIFF
--- a/pyworkflow/em/packages/grigoriefflab/viewer.py
+++ b/pyworkflow/em/packages/grigoriefflab/viewer.py
@@ -28,8 +28,9 @@ Visualization of the results of the Frealign protocol.
 """
 import os
 from os.path import exists, relpath
-from pyworkflow.utils.path import cleanPath
+from pyworkflow.utils.path import cleanPath, removeBaseExt
 from pyworkflow.viewer import (ProtocolViewer, DESKTOP_TKINTER, WEB_DJANGO, Viewer)
+from pyworkflow.em.viewer import CtfView
 import pyworkflow.em as em
 import pyworkflow.em.showj as showj
 from pyworkflow.em.plotter import EmPlotter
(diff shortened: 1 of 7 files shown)


azazellochg:
@delarosatrevin , i have tested **Analyze results** for several imports as well as ctffind3/4 and xmipp-ctf protocols. _Display CTF fitting_ works in all cases, where ctffind4 was used - that's OK. But _Show CTF profile_ works only for ctffind3/4 or xmipp-ctf protocol. _Show CTF profile_ for all import cases crashes with same java error.

I will send you the folder with 3 test mics and associated ctfs as well as workflow I used, so you can check it.

delarosatrevin:
That would be helpful, the mics/ctfs and workflow to repeat.

delarosatrevin:
I found the difference in your case between the imported CTF and the computed ones. The problem was not related to import but to the PSD size, the one from the imports was 1024 and the computed one by default is 256. Now I always read the PSD with 256 pixel size. 

Anyway I was not getting an Exception here after my previous changes, the CTF Profile GUI just went a bit crazy. Remember to recompile after the update since the changes are in Java code that requires re-compile.

Let me know if works now for you.

azazellochg:
You were right, I forgot to recompile java code, sorry. Everything works now!

azazellochg:
@delarosatrevin , one small detail: if I have only ctffind3 log files and I import CTFs, on the place of PSD file (which should be a red cross since it is missing) there is a micrograph displayed. Do we leave it like that or need to correct?

delarosatrevin:
I think this is a side effect that we are trying to render a column but it does not exists (psd image) and then render the next column available (micrograph). I would say that it is better to show the filename and not render the micrograph because can take a time if there are big micrographs. Anyway this is not a big issue and we could stop here. We could add a new issue to solve this in the next version.

PR #351, opened 2016-03-02
Fix output coordinates

Vilax:
Problems in Summary and methods in protocols that involve tilt pairs have been fixed


delarosatrevin:
We will do a single PR from v1.0 to master for release v1.0.1, so this is not necessary.

PR #352, opened 2016-03-02
Fix output coordinates

Vilax:
Summary and methods in tilt pairs protocols have been fixed


delarosatrevin:
@Vilax , you should remove some debugging prints that are still in the code. You can also remove some commented blocks that are not longer used.

PR #364, opened 2016-03-09
don't overwrite LD_LIBRARY_PATH

pescobar:


install/funcs.py
```diff
@@ -476,7 +476,7 @@ def path(x):
 
         environ = {
             'PYTHONHOME': prefix,
-            'LD_LIBRARY_PATH': '%s/lib' % prefix,
+            'LD_LIBRARY_PATH': '%s/lib:%s' % (prefix, os.environ.get('LD_LIBRARY_PATH')),
```
delarosatrevin: Just to be safe when LD_LIBRARY_PATH is not defined in the environment, we should use:
os.environ.get('LD_LIBRARY_PATH', '') 
if not defined, this will return the empty string. I'm not sure if will work in csh shell, where I think /path/:  is not valid, but maybe Python take cares of this.

pescobar:
@delarosatrevin when the var doesn't exists python returns `None`. Isn't that fine? Anyway I did your suggested change

```
In [3]: print os.getenv('asdfadfadfadsfasdf')
None
```

delarosatrevin:
You are right, if None, it will not fail, but you will end with a LD_LIBRARY_PATH like:
'/scipion/software/lib:None'  which is not wrong, but may be confusing. 

Thanks for the change.

PR #388, opened 2016-04-14
Fix dimred

cossorzano:


delarosatrevin:
Not much to say here.

juannavascalvente:
Dimred tests passed.

PR #389, opened 2016-04-19
Implemented new protocol to remove a projection image(s) from an experimental image(s)

josuegbl:


delarosatrevin:
I have two comments, one more important that the other:
* First, there is NO TEST for the protocol....this is a MUST for adding a new protocol to devel.
* Second, and just a minor thing, I would suggest to add a parameter to --minus (such as adjust) instead of adding another option (--minusAdjust, which by the way does not complain the params convention).

PR #393, opened 2016-04-21
relion bug related with generation of SetOfParticles from an intermediate iteration fixed.

josuegbl:
test of relion refine updated.
fixed summary for  relion postprocess.


delarosatrevin:
Reviewed (without passing Relion tests)

PR #394, opened 2016-04-24
modifying input subsets related code to avoid making error for useres

mohsenkazemi:
now, the protocol take care if the input subsets are greater than 1/2 of set of input particles (instead of making Error message for the user)

### DIFF
--- a/pyworkflow/em/packages/xmipp3/protocol_validate_overfitting.py
+++ b/pyworkflow/em/packages/xmipp3/protocol_validate_overfitting.py
@@ -72,9 +72,9 @@ class XmippProtValidateOverfitting(ProtReconstruct3D):
                       help='Select the input images from the project.')     
         form.addParam('doResize', BooleanParam, default=False,
                       label='Resize input particles and volume?',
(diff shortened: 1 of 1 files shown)


pyworkflow/em/packages/xmipp3/protocol_validate_overfitting.py
```diff
@@ -83,20 +83,26 @@ def _defineParams(self, form):
         form.addParam('input3DReference', PointerParam,
                  pointerClass='Volume,SetOfVolumes',
                  label='Initial 3D reference volume', 
-                 help="Input 3D reference reconstruction to"
+                 help="Input 3D reference reconstruction to "
                       "alignment gaussian nise.")
```
delarosatrevin: I guess there is a typo here: nise -> noise?....and not sure if 'to align' instead of 'to alignment'

mohsenkazemi: done!
Tnx J. Miguel

PR #401, opened 2016-04-26
new developed protocol to combine two pbd files

mohsenkazemi:


delarosatrevin:
One of the requirements to include a protocol into devel is to create at least one test for it.
I would recommend that you test here the different options the protocol allows (from PDB_ID, file or object).
You can take a look at other PDB-related tests by typing:
./scipion test --grep pdb --show

pyworkflow/em/packages/xmipp3/pdb/protocol_combine_pdb.py
```diff
@@ -0,0 +1,195 @@
...
+        pdbFn2 = self._getPdbFileName2()
+        fnOut = self._getPdbOutName()
+        
+        self.info("The 1st input file: " + pdbFn1)
+        self.info("The 2nd input file: " + pdbFn2)
+        self.info("Output file: " + fnOut)
+        
+        fh = open(fnOut, "a")
+        for line in open(pdbFn1,"r"): 
+            list = line.split()
```
delarosatrevin: Here (and in the next loop) the 'list' variable seems unnecessary since you could directly take the id.
Apart from that, both 'list' and 'id' are bad names for variables in python since they are reserved words
for the systems.

PR #407, opened 2016-04-28
Imagic msa

delarosatrevin:
* Added package for IMAGIC protocols (MSA workflow)
Some bugfixes:
* Fixed bug in apply alignment using matrix (Python binding)
* Fixed bug in Spider (now applying alignment if exists)
* Back-ported (from scipion-box) some utilities to use EMAN for some image conversions.


pyworkflow/tests/em/data/test_data.py
```diff
@@ -86,6 +87,15 @@ def testExistLocation(self):
...
         self.assertTrue(ih.existsLocation(newFn))
 
+    def test_convertMicrographs(self):
+        """ Convert micrograhs to different formats.
+         EMAN2 required for .img
+        """
+        micFn = self.dataset.getFile('micrographs/BPV_1386.mrc')
+        outFn = pwutils.replaceBaseExt(micFn, 'img')
+        ih = ImageHandler()
+        ih.convert(micFn, os.path.join('/home/josem', outFn))
```
josuegbl: directory /home/josem/ not exists always. The path to do the convert must be more general

pyworkflow/tests/em/data/test_data.py
```diff
@@ -49,6 +49,7 @@ def testMd(self):
         self.assertEqual(xList,x)
         self.assertEqual(yList,y)
```
josuegbl: thre are any test for .dm4 format? It seems the test not view this condition.

josuegbl:
Review the comments in the code.

delarosatrevin:
@josuegbl , check again and pass the added tests.

josuegbl:
everything ok.... imagehandler tests passed.

Issue #56, opened 2015-06-26
spider refinament does not create any output


rmarabini:

rmarabini:
No symmetry is asked in spider refinement!

azazellochg:
the protocol crashes at the very beginning, when running [prepare.pam](scipion/pyworkflow/em/packages/spider/scripts/projmatch/Refinement/prepare.pam) script:

```
  -- START OF: prepare.pam    --
 .OPERATION: [iter]=[iter1]  ; Needed for naming initial files
 .OPERATION: VM  
 .SYSTEM COMMAND: 
     mkdir -p work/ final/ input/
 .OPERATION: CP  ; Copy volume
 .INPUT FILE NAME OR TEMPLATE (E.G. STK@****):  ../Reconstruction/vol01 
 *** FILE NOT FOUND: ../Reconstruction/vol01.stk
 *** ERROR: OPENFIL; FILE NOT FOUND: ../Reconstruction/vol01.stk 
```

azazellochg:
Also, spider refinement protocol fails if ctf estimation was not done, since it needs defocus groups

delarosatrevin:
@azazellochg this is an easy one, we could put a condition in the inputParticles to allow only the ones which haveCTF.

azazellochg:
fixed in cd443bc

PR #410, opened 2016-04-29
Fixed several spider bugs

delarosatrevin:


josuegbl:
Selection of classes2D, from classify ward protocol, fails.

azazellochg:
I would suggest to backport spider-related fixes from devel to this branch (into 1.0.1 eventually), since we fixed there alignment parameters transfer and stack writing.

delarosatrevin:
@azazellochg , Totally agree, we should back-port those fixes. We need to be careful since there is a modification in the ImageHandler class, we need to be sure not to break something else.

azazellochg:
@delarosatrevin , I have done it however you might want to re-check the last changes after me.  Also, commit cd443bc closes #56

delarosatrevin:
Thanks a lot @azazellochg ! I have looked carefully and the back-port seems correct.
I also passed some basic tests and others related to Spider (all finished correctly).

I will keep this PR opens until and can fix the remaining issues.

azazellochg:
In _Spider align ap sr_ protocol we should remove centering option - it is not used. Centering is done using a low-pass filtered disc. Or, we could modify original Spider apsr4class.msa script.

delarosatrevin:
Last commits fix all issues mentioned in #411.
@azazellochg , could you double-check and approve the merge?

delarosatrevin:
I'm adding a last change, to remove classes with only one child in the display tree.

azazellochg:
@delarosatrevin , great work here! all issue look fixed to me. What do you think about centering option in ap sr?

delarosatrevin:
If the centering option is not used, I think the best option right now is to remove it from the protocol (I can do that)...since @taputechnic have updated some of the scripts, maybe we should plan an update for the 1.1 release.

PR #435, opened 2016-05-13
editing the legend in viewer and forcing thread to 1 for reconstruct fourier in main protocol

mohsenkazemi:


pyworkflow/em/packages/xmipp3/protocol_validate_overfitting.py
```diff
@@ -114,7 +114,7 @@ def _defineParams(self, form):
                       expertLevel=LEVEL_ADVANCED,
                       label="Angular sampling rate")  
                       
-        form.addParallelSection(threads=4, mpi=1)
+        form.addParallelSection(threads=1, mpi=1)
```
delarosatrevin: If we are going to disable threads from this protocol, we should use threads=0 to remove it even from the protocol form GUI.

pyworkflow/em/packages/xmipp3/protocol_validate_overfitting.py
```diff
@@ -114,7 +114,7 @@ def _defineParams(self, form):
                       expertLevel=LEVEL_ADVANCED,
                       label="Angular sampling rate")  
                       
-        form.addParallelSection(threads=4, mpi=1)
+        form.addParallelSection(threads=0, mpi=1)
```
delarosatrevin: Maybe mpi=4 (or even more)? What do you think is a reasonable default value for using this protocol?

PR #440, opened 2016-05-20
PDFReport created

cossorzano:


cossorzano:
For some reason I cannot see, the protocol is recognized as xmipp3-pdf report instead of scipion-pdf report

PR #462, opened 2016-06-06
Fixing race conditions when storing return value from sql functions

juannavascalvente:
The static variable "int MDSql::rc;" is shared by different functions in the metadata_sql.cpp file. This could produce race conditions when several threads execute SQL sentences and overwritting this return variable.

Some of the functions do not really need to use this variable. The new version does not store the return value if not needed and keeps this variable when using a static member function.


delarosatrevin:
I took a look into the code and seems fine. @juannavascalvente, after you pass the tests in this branch, merge the PR.

PR #477, opened 2016-06-13
Remove memory leaks in projection matching when allocating threads data

juannavascalvente:
The thread were allocating memory into a for loop without deallocating the memory. Now it is allocated just once and deallocated when the threads exit.


software/em/xmipp/libraries/reconstruction/angular_projection_matching.h
```diff
@@ -134,14 +134,12 @@ class ProgAngularProjectionMatching : public XmippProgram
     /** Threads */
     int threads;
     //Numbre of possible orientations
```
delarosatrevin: There is a typo here

delarosatrevin:
@juannavascalvente , this is a serious memory leak. Can you check if it is also in v1.0 branch?
The threads allocated memory is freed?
@rmarabini, can you take a look here?

juannavascalvente:
This bug is also in v1.0.

delarosatrevin:
@juannavascalvente , please fix it in v1.0, check with @rmarabini that everything is correct...and we can create a new PR into v1.0. All the bug fixed there are merged later into devel.

rmarabini:
If I am not wrong numOrientations is a parameter introduced by @jvargas.

435b1cfe (jvargas                       2015-09-09 10:03:20 +0200  115)     addParamsLine("  [--number_orientations <numOrientations=1>]  : Number of possible orientations for each experimental image");

Question for @jvargar. Are you interested in this option or should we restore the previous projection matching version?

cossorzano:
I am also using this option in Highres. I would like to keep it

PR #480, opened 2016-06-14
Devel update 20160614

delarosatrevin:
Update the 'devel' branch with the bugfixes from 'v1.0'.

@josuegbl , I created a branch to update devel with latest bug-fixes but I'm surprise of so few changes.  Have you done this in the recent days?


josuegbl:
This is so because @cossorzano did a merge from V1.0 to devel branch directly today. I'm surprised because devel branch is not protected to avoid these behaviors. I check this changes and merge the pull request.

delarosatrevin:
Ok, got it. I will check the configuration.

PR #482, opened 2016-06-15
Create gallery of projections added as a protocol

cossorzano:


delarosatrevin:
@cossorzano , Missing a test for this protocol. You could any volume from the exiting test data and add a couple test cases. Could you improve a little bit the protocol and parameters help? Sorry for bothering, but users will appreciate it.

cossorzano:
Test and better help added

delarosatrevin:
@cossorzano , I did some format changes...moreover, I grouped rot and tilt parameters into a line with Mix, Max and Step. If you don't like it we can revert this change. I also remove the EXPERT_LEVEL, since there are not much parameter I think users will usually like to touch those ones. You can comment about with @josuegbl  or @rmarabini ...and again, if you don't like it we can revert it. I added a foolish test check and an extra case.

PR #487, opened 2016-06-21
Check if PID is alive for running jobs when refreshing the desktop

### LINKED ISSUE #183: Check if PID is alive for running jobs when refreshing the desktop

pconesa:
Invokes refresh runs the first time a project window is loaded and on manual refresh.

Closes #183


delarosatrevin:
@pconesa , I have done some tests in my laptop and seems to be working.
I think we need to do some further tests in crunchy when it become available.

PR #514, opened 2016-07-10
add motioncor2 program

azazellochg:
Beta version can be found [here](http://msg.ucsf.edu/em/software/motioncor2.html)


delarosatrevin:
Dear @azazellochg, thanks a lot for adding support for motioncorr2. I have two important points that I want to to address here:
* First, It would be better to merge this PR into the scipion-box branch since we have refactored the movie protocols there and we are in the process to merge it to the devel. The movie protocols for alignment now inherits form a base class so some parameters are defined there and the good thing is that automatically will be available for the streaming processing. The output from this protocols have also been more standardized.
* Second, it would be better to add this program inside the existing package **motioncorr** and not create a new one. We can handle supported versions inside the package (look at eman or relion). The other approach, if we want to allow both programs, to follow an approach similar to CTFFIND, but there is only one protocol that allows to execute both. 

We can further discuss about this and work together.

azazellochg:
Hi @delarosatrevin ,
I have noticed that dosefgpu files in 1.0.1 are quite out-of-date and thought that you plan to put all movie protocols into "xmipp movie alignment", but now I see that you have separated xmipp algorithms (op.flow, correlation, average) from motioncorr in **devel-scipion-box** branch. I completely agree with this approach, since more movie alignment programs are coming (I know at least about 2 more in development) it's good not to put them all in one protocol.

So, I guess I will rather look at **devel-scipion-box** branch and push there into motioncorr package.

PR #522, opened 2016-07-15
Co small fixes

cossorzano:
Jose Miguel, please, do not delete my branch after merging

### DIFF
--- a/pyworkflow/em/packages/atsas/protocol_pdb_to_saxs.py
+++ b/pyworkflow/em/packages/atsas/protocol_pdb_to_saxs.py
@@ -26,45 +26,57 @@
 
 import math
 from glob import glob
+import numpy
 
 from pyworkflow.em import *  
 from pyworkflow.utils import * 
-from pyworkflow.protocol.constants import LEVEL_ADVANCED, LEVEL_ADVANCED
+from pyworkflow.protocol.constants import LEVEL_ADVANCED
 import atsas
 from pyworkflow.utils.path import createLink
 
 # TODO: Move to 3D Tools
 class AtsasProtConvertPdbToSAXS(ProtPreprocessVolumes):
-    """ Protocol for converting a PDB file (true atoms or pseudoatoms) into a SAXS curve.
+    """ Protocol for converting a PDB file (true atoms or pseudoatoms) into a
+    SAXS curve.
     
-    This is actually a wrapper to the program Crysol from Atsas ( see documentation at http://www.embl-hamburg.de/biosaxs/manuals/crysol.html ) """
+    This is actually a wrapper to the program Crysol from Atsas.
+    See documentation at:
+       http://www.embl-hamburg.de/biosaxs/manuals/crysol.html
+    """
     _label = 'convert PDB to SAXS curve'
     
-    #--------------------------- DEFINE param functions --------------------------------------------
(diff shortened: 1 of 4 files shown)


pyworkflow/em/packages/xmipp3/protocol_multiple_fscs.py
```diff
@@ -107,13 +107,12 @@ def prepareReferenceStep(self,volId):
...
 
     def compareVolumeStep(self, volLoc, i):
-        ih = em.ImageHandler()
         fnRef = self._getExtraPath("reference.vol")
         sampling = self.referenceVolume.get().getSamplingRate()
         fnRoot = self._getExtraPath("volume_%02d" % i)
         fnVol = fnRoot + ".vol"
-        ih.convert(volLoc, fnVol)
-
+        self.runJob("xmipp_image_convert","-i %s -o %s -t vol"%(volLoc,fnVol))
```
delarosatrevin: @cossorzano , Why do you prefer to use runJob("xmipp_image_convert") instead of ImageHandler.convert, which is the recommended way?

pyworkflow/em/packages/atsas/protocol_pdb_to_saxs.py
```diff
@@ -94,6 +94,16 @@ def _summary(self):
             summary.append('Experimental SAXS curve: %s'%self.experimentalSAXS.get())
         if not self.otherCrysol.empty():
             summary.append('Other crysol parameters: %d'%self.otherCrysol.get())
+
+        # Goodness of fit
+        fnInt=self._getPath("pseudoatoms00.fit")
+        if exists(fnInt):
+            import numpy
```
delarosatrevin: Import of numpy should not be here but on the top of the file. This is only done in a few exceptional cases where we want to delay the dependency, but it is not the case of numpy.

pyworkflow/em/packages/xmipp3/protocol_multiple_fscs.py
cossorzano: Because it was failing when the input is a volume .mrc produced by Relion

pyworkflow/em/packages/atsas/protocol_pdb_to_saxs.py
cossorzano: I think it really is the case, because it we re using numpy only at the 
summary

delarosatrevin: But numpy is a system-wide dependency, so there is no need to import here...the general case is to import at the top.

pyworkflow/em/packages/xmipp3/protocol_multiple_fscs.py
delarosatrevin: OK, but for the future we should address this issue in a more general way the ImageHandler could do it....it is a bit overkilling to spawn new processes just to convert a volume.

cossorzano: I agree, but in its current form it is a bug of ImageHandler that we 
have to circumvent

delarosatrevin:
@cossorzano , I have done some code formatting in the SAXS to PDB protocol to meet our conventions. Could you check that I have not broke anything? (since this protocol does not have any test I can not do it myself)

PR #528, opened 2016-07-21
Check pid for protocol status

### LINKED ISSUE #183: Check if PID is alive for running jobs when refreshing the desktop

### LINKED ISSUE #511: Delete key in seaching box
When you want to delete the text in the search box using the delete key, the selected protocol is attempted to be deleted

issue

Regards

Vilas

pconesa:
Now it's checking properly the PID.
Also check only against running status, leaving outside the rest active ones(launched, interactive).
Additionally: Do not try to remove a protocol if delete is pressed while on the search box (the one in the logs tab).

Closes #183 
Closes #511 

Use \u format for unicode special characters, since the characters it self  generates compiling error in javac when the locales are not set properly.

### DIFF
--- a/pyworkflow/gui/project/viewprotocols.py
+++ b/pyworkflow/gui/project/viewprotocols.py
@@ -513,6 +513,7 @@ class ProtocolsView(tk.Frame):
     This extended tk.Frame is what will appear when Protocols is on.
     """
 
+    RUNS_CANVAS_NAME = "runs_canvas"
     def __init__(self, parent, windows, **args):
         tk.Frame.__init__(self, parent, **args)
         # Load global configuration
@@ -934,7 +935,7 @@ class ProtocolsView(tk.Frame):
     def createRunsGraph(self, parent):
         self.runsGraphCanvas = pwgui.Canvas(parent, width=400, height=400, 
                                 tooltipCallback=self._runItemTooltip,
-                                tooltipDelay=1000)
+                                tooltipDelay=1000, name=ProtocolsView.RUNS_CANVAS_NAME, takefocus=True)
 
         self.runsGraphCanvas.onClickCallback = self._runItemClick
         self.runsGraphCanvas.onDoubleClickCallback = self._runItemDoubleClick
@@ -1275,6 +1276,8 @@ class ProtocolsView(tk.Frame):
         
     def _runItemClick(self, item=None):
 
+        self.runsGraphCanvas.focus_set()
+
         # Get last selected item for tree or graph
         if self.runsView == VIEW_LIST:
             prot = self.project.mapper.selectById(int(self.runsTree.getFirst()))
@@ -1582,7 +1585,7 @@ class ProtocolsView(tk.Frame):
         widget = self.focus_get()
 
         # If it's the search box
-        if str(widget).endswith(SEARCH_BOX_NAME):
+        if not str(widget).endswith(ProtocolsView.RUNS_CANVAS_NAME):
             return
 
         protocols = self._getSelectedProtocols()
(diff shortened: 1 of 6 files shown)


pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1576,6 +1577,14 @@ def _continueProtocol(self, prot):
         self._scheduleRunsUpdate()
         
     def _deleteProtocol(self):
+
```
delarosatrevin: @pconesa , For this I will suggest to go the other way around. Name the canvas widget with the protocol boxes and only if the name is that one, react on the 'delete' event. Because imagine that we add in a future another dialog with edition, then it will suffer the same issue.

pyworkflow/protocol/protocol.py
```diff
@@ -967,6 +967,7 @@ def run(self):
         
         self.info(greenStr('RUNNING PROTOCOL -----------------'))
         self._pid.set(os.getpid())
+        self.info('          PID: %s' % self._pid)
```
delarosatrevin: I think this is not relevant for the user. Maybe we could log it only if the SCIPION_DEBUG flag is active.

software/em/xmipp/java/src/xmipp/viewer/models/MetadataTableModel.java
```diff
@@ -483,13 +483,15 @@ public void mouseClicked(MouseEvent e) {
 
             final String columnName = data.labels.get(sortColumnIndex).labelName;
 
-            column.setHeaderValue( "⌛ " + columnName);
+            //column.setHeaderValue( "⌛ " + columnName);
+            column.setHeaderValue( "\u231B " + columnName);
```
delarosatrevin: Add a comment about this change also in the PR.

pconesa: Done, forgot that change.

pyworkflow/protocol/protocol.py
pconesa: It could be useful if a process ends unexpectedly, I think it doesn't harm. I left it intentionally. But will do it if you insist.

pyworkflow/gui/project/viewprotocols.py
pconesa: I know this is a bit dirty, but the canvas never gets the focus. I'll have a look at it, since you made a good point.

I'll have a look at this:

takefocus=
Indicates that the user can use the Tab key to move to this widget. Default is an empty string, which means that the canvas widget accepts focus only if it has any keyboard bindings. (takeFocus/TakeFocus).

Which now seems to be inactive.

delarosatrevin: OK, take a look just if there is any easy way to overcome this, if not, we can live with it.

pyworkflow/protocol/protocol.py
delarosatrevin: I agree that it does not harm, we can keep it.

delarosatrevin:
@pconesa , I made some minor changes just to clean up a bit not needed/used code. I have done a few tests on the checking and seems to work. The only downside is the black border that now appears/disappears in the canvas when get/release the focus...weird stuff from tkinter. Check if you can remove that border easily, if not is also ok. After that you can merge the PR.

pconesa:
highlightthickness=0 does the trick

delarosatrevin:
Great, commit that change to test it and merge the PR.

delarosatrevin:
Ups, I just commented while you were pushing the change.

PR #534, opened 2016-07-22
Co small fixes

cossorzano:


pyworkflow/em/packages/xmipp3/protocol_multiple_fscs.py
```diff
@@ -74,7 +74,7 @@ def _insertAllSteps(self):
         allVols = []
         for i, vol in enumerate(self.inputVolumes):
             volId = self._insertFunctionStep('compareVolumeStep',
-                                             vol.get().getLocation(), i+1,
+                                             vol.get().getFileName(), i+1,
```
delarosatrevin: This will not works if the volume comes is inside a stack. 
You you pass the location as before and use the function locationToXmipp to get the number@filename that can be used with xmipp_image_convert.

PR #540, opened 2016-07-27
Fix intro tutorial

azazellochg:


delarosatrevin:
I merged this fix in the devel branch in #545
We do not plan to update the v1.0 branch.

PR #544, opened 2016-07-28
fix workflow_betagal1.json

azazellochg:
a minor correction. Betagal dataset is 3.54A/px from 14um Falcon II, giving 39548 magnification


delarosatrevin:
I merged this fix in the devel branch in #545
We do not plan to update the v1.0 branch.

PR #554, opened 2016-08-02
az small fixes

azazellochg:


delarosatrevin:
@azazellochg , you do not need to create a PR to commit changes to *devel-fixes* branch. The idea of the branch is just to make these small changes more agile. When we accumulate some changes we then create the PR from *devel-fixes* to *devel*. Moreover, I'm the only one that have access to your scipion fork, so if someone else need to change something, will be better to work in *devel-fixes* directly.

azazellochg:
Alright, I will push directly next time.

delarosatrevin:
Great!

Issue #546, opened 2016-07-29
eman2 - initial model: results viewer improvements


azazellochg:
Can be improved by displaying not just volumes, but [also](http://blake.bcm.edu/emanwiki/EMAN2/Programs/e2initialmodel#Results) reprojections+class-averages file, final reprojections file and initial model file. All this files are produced by the protocol and are in _extra/initial_model_ folder.

delarosatrevin:
After updating, you should have the following:

![image](https://cloud.githubusercontent.com/assets/8441688/17486270/824f2de4-5d90-11e6-935b-bb352bf37a61.png)

azazellochg:
@delarosatrevin Great, looks perfect!

PR #560, opened 2016-08-08
Re-factored how Object-Commands are handled

delarosatrevin:
Object-Commands are commands passed to Showj Java app when visualizing some to dynamically add command available for specific items. Then Showj call back the project gui that is listening at some port and react to this commands. 

Before, the gui/project.py contains the dirty includes for the commands. Now it has a **registerCommand** method to register new command without changing the code at this levels.

After this, I added 3 extra commands to improve Eman2 initial-volume visualization as suggested by @azazellochg  in #546 

@azazellochg , could you double check that everything is working as expected?


azazellochg:
Everything looks fine!

PR #561, opened 2016-08-09
Sort objects in browse window

### LINKED ISSUE #530: sort objects by creation date in browse window
In version 1.0.1:
I think it would be nice if the objects in browse window were sorted by creation date (creation column) by default with a possibility to sort by name (object column). Otherwise it becomes difficult to find something if I have >30 different sets

error3

pconesa:
Treeview invokes sorting if provider allows it. 
FileTreeProvider implements sorting.
Closes #530


delarosatrevin:
Nice work @pconesa !!!

Before merging we should implement the sorting for: 

* *Browse objects* (SubclassesTreeProvider in form.py, also check other providers there)
Indeed, this one was the original request.
image

* *Runs list* (RunsTreeProvider in viewprotocols.py)
image

pconesa:
@delarosatrevin , all done.

PR #562, opened 2016-08-09
relion particle sorting protocol

azazellochg:
It looks like the protocol is almost ready, except couple of things left to fix:

* I still had to create a [writeReferencesNew](https://github.com/I2PC/scipion/blob/7a028608b8e8add394f277fa0ff34a48bc28fdbf/pyworkflow/em/packages/relion/convert.py#L579) function in relion - convert.py (old one is used only in relion autopicking), since the old one created reference star file with [relative](https://github.com/I2PC/scipion/blob/7a028608b8e8add394f277fa0ff34a48bc28fdbf/pyworkflow/em/packages/relion/convert.py#L568) image names (x@input_refs.stk),  and I needed x@Runs/.../extra/input_refs.stk. @delarosatrevin , maybe you could suggest how to fix this and simply update the old function.

* Also, do we need a test for this protocol?


pyworkflow/em/packages/relion/protocol_sort.py
```diff
@@ -0,0 +1,257 @@
...
+from pyworkflow.protocol.params import (PointerParam, FloatParam, StringParam,
+                                        BooleanParam, IntParam, LEVEL_ADVANCED)
+from pyworkflow.utils import removeExt
+from pyworkflow.em.protocol import ProtParticles
+from convert import (convertBinaryVol, readSetOfParticles,
+                     writeSetOfParticles, writeReferencesNew)
+import pyworkflow.em.metadata as md
+
+
+class ProtRelionSort(ProtParticles):
```
delarosatrevin: Could we called it **ProtRelionSortParticles** ? (Just in case there is any other sort protocol in the future)

azazellochg: @delarosatrevin sure!

pyworkflow/em/packages/relion/convert.py
```diff
@@ -561,11 +569,70 @@ def writeReferences(inputSet, outputRoot):
...
-            
+
     elif isinstance(inputSet, em.SetOfClasses2D):
         pass
     else:
-        raise Exception('Invalid object type: %s' % type(inputSet)) 
+        raise Exception('Invalid object type: %s' % type(inputSet))
+
+
+def writeReferencesNew(inputSet, outputRoot, **kwargs):
```
delarosatrevin: It is very easy to modify **writeReference** function to set the filename in the star file either relative or with the project path. Anyway you have added other possible input types, do you plan to use them in other protocol? Or was just to make more complete the function?

pyworkflow/em/packages/relion/protocol_sort.py
```diff
@@ -0,0 +1,257 @@
+# **************************************************************************
+# *
+# * Authors:     Grigory Sharov     (sharov@igbmc.fr)
+# *
+# * Unidad de  Bioinformatica of Centro Nacional de Biotecnologia , CSIC
```
delarosatrevin: L'Institut de genetique et de biologie moleculaire et cellulaire (IGBMC) ?

pyworkflow/em/packages/relion/convert.py
azazellochg: yes, they are used in the protocol. Particles can be sorted after 2d/3d classification , 3d refinement or autopicking

pyworkflow/em/packages/relion/viewer.py
```diff
@@ -1400,3 +1402,50 @@ def _getLegend(self, label):
...
+    def visualize(self, obj, **args):
+        self._visualize(obj, **args)
+
+        for v in self._views:
+            v.show()
+
+    def _visualize(self, obj, **args):
+        cls = type(obj)
+
+        if issubclass(cls, SetOfParticles):
```
delarosatrevin: You don't need to handle also the case of **SetOfParticles** and visualize directly in the other case.
I can change this later.

delarosatrevin:
@azazellochg , I have made some comments in the code. Apart from that, as you said, it would be good to add a test to this protocol. Maybe we could import some particles that comes directly from Relion and have the FOM information? If not, I guess we should run a picking first, which will take more time for running this test.

pyworkflow/em/packages/relion/protocol_sort.py
delarosatrevin: OK, I'm looking at it, I can change it.

pyworkflow/em/packages/relion/convert.py
delarosatrevin: OK, I will rename your function as the writeReferences and modify it to handle the paths in one way or another.

delarosatrevin:
@azazellochg , I have been looking into the protocol and I have some extra questions for you.
I notice that we have two main inputs: *inputParticles* and *inputReferences* but in some cases are redundant, since when we have *SetOfClasses2D/3D*, we have the associated particles. Moreover, the case of *Volume* it seems that is not used.

azazellochg:
@delarosatrevin indeed. I was thinking a lot on protocol design since in case after auto-picking you have to have extracted particles with FOM info etc. However, in case after 2d/3d classification or refinement you can provide a set of 2d/3d classes or a volume as a reference and a set of particles can be found from sourceRelation. Maybe we could add a parameter to the protocol form: if input is extracted particles (after autopicking), then user has to provide particles and references himself, and if input is after classification/refinement only setofclasses or a volume is enough.

pyworkflow/em/packages/relion/protocol_sort.py
azazellochg: sure, you can change to that ^)

pyworkflow/em/packages/relion/protocol_sort.py
```diff
@@ -0,0 +1,260 @@
...
+        sortedImgSet.copyInfo(imgSet)
+        readSetOfParticles(self._getFileName('output_star'),
+                           sortedImgSet,
+                           alignType=imgSet.getAlignment())
+
+        self._defineOutputs(outputParticles=sortedImgSet)
+        self._defineSourceRelation(imgSet, sortedImgSet)
+
+    #--------------------------- INFO functions --------------------------------
+    def _validate(self):
```
azazellochg: Todo: add validation that input particles have rlnClassNumber, at least rlnPsiAngle and in-plane shifts. FOM is not used for sorting protocol (my mistake, I forgot).

azazellochg:
I created a test protocol, right now it needs some extra files in relion_tutorial test data. Also I could not create a test data for autopicking case since relion autopick protocol is broken right now in this branch.

azazellochg:
Now auto-picking works again, also updated the test. Now sorting works after cl2d and picking/extraction, but not after cl3d/ref3d.

delarosatrevin:
OK, I will check all with my changes.

pyworkflow/em/packages/xmipp3/convert.py
```diff
@@ -56,6 +56,13 @@
              ("_y", xmipp.MDL_YCOOR) 
              ])
 
+COOR_EXTRA_LABELS = [
```
delarosatrevin: @azazellochg Why the Relion coordinate labels are needed here? Without that, it is not propagated the Coordinate information into the particle? Maybe we should fix that in a more general way.

azazellochg: @delarosatrevin yes, without this SetOfCoordinates does not store autopicking metadata from relion

delarosatrevin: Hum, as I said, we need to make this more general, imagine if in the future other information (from Relion or not) comes in the Coordinate, we should propagate it.

delarosatrevin:
@azazellochg , I finally finished the changes in the protocol. Now I think that it is more simple for the user, but complicated a little bit for the developer. Now the main input is:

inputSet: SetOfClasses2D/3D, or SetOfParticles (from auto-pick or auto-refine)
referenceAverage -> only in the auto-pick case
referenceVolume -> only in the auto-refine case

Please check if I missed some detail. It is also fine if you think the other way is better and we change back to that way. Check also the class mapping with the new changes. I updated the test properly and all are passing (I also uploaded the new test files)

delarosatrevin:
Moreover, I did some modifications to the libraries that I would like to summarize:

- Now the *SetOfClasses* has two new functions:
   - iterClassItems : to iterate over all items of all classes (e.g., all particles in classes 2d)
   - iterRepresentative: to iterate over all representative of classes (e.g., all averages)
- Now conditions in the protocols parameters are evaluated in a better way. We now support functions or input objects usage in conditions.

azazellochg:
@delarosatrevin great work! I had a look and there are few things left to fix:

- [ ] particle diameter wizard is not working
- [ ] in metadata table view for particles we should show Zscore column by default
- [ ] in the viewer we should remove target SetOfImages, otherwise scipion tries to open any set of anything with relionsortviewer

delarosatrevin:
@azazellochg , Wizard working, I set that if coming from classes, the wizard displays the averages or volume middle slice, you can easily change this by removing a couple of lines in *RelionSortMaskWizard._getProvider* method and always show input particles.

azazellochg:
i think it is ready for merge

PR #563, opened 2016-08-10
Adding labels to scipion script to filter tests

juannavascalvente:
The scipion script has been modified to accept a new parameter "--label" that filters the test to be listed by "--show" or executed by "--run". This new label parameter accepts a list of labels and filters the test that match the set of labels.

Currently the test accept three label values: pull, daily and weekly. It defines the execution of the test in buildbot: pull when it must be executed under a new pull request, daily to be executed once a day and weekly to be executed during the weekend. 

At this moment, only one TestCanvas is defined as pull request, and all the others are set to weekly.


juannavascalvente:
The scipion xmipp basic test are now discovered when using the --label parameter in scipion. Those tests are coded in C++ and so they do not have a label and cannot be classified as in python. Now they are "classified" automatically as "pull" label in the "_visitTests" procedure by calling the "discoverXmippTest" procedure.

PR #566, opened 2016-08-15
refactor xmipp extract particles protocol

### LINKED ISSUE #313: Allow downsample when extracting particles even if using same micrographs
The downsampling parameters make sense even if using the same set of micrograph to extract the particles.

### LINKED ISSUE #543: Xmipp Extract particles protocol should clean up tmp files at run time
Xmipp Extract particles generates a bunch of intermediate files that are only deleted when run has finished. This could cause problems of lacking of disk space.

### LINKED ISSUE #569: ProtImportMicrographsTiltPairs does not set micName
ProtImportMicrographsTiltPairs does not set micName in the output set

azazellochg:
- [x] Now particles can be downsampled during extraction (closes #313)
- [x] Removed outdated options and unused functions, some code formatting
- [x] Remove temporary mics after extraction (closes #543)
- [x] fix all related tests/workflows
- [x] refactor extract particle pairs protocol
- [x] fixed #569 bug when micName was not set

Do we need to keep downsampled particle stack in a separate file (right now extracted particles are re-written)?

TODO:
- [x] check all modified tests/workflows


azazellochg:
@delarosatrevin i discovered that after tilt pair picking protocol neither _coordinates(un)tilted_ nor _coordinates_pairs_ (main output set) database contains **micName** (column is empty), is it a bug? Only **micId** is present

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
```diff
@@ -277,76 +268,86 @@ def writePosFilesStep(self):
...
+            if self.doDownsample:
+                downFactor = self.downFactor.get()
+                self.runDownsample(outputRoot + '.stk', boxSize, downFactor)
+
             # Normalize 
             if self.doNormalize:
                 self.runNormalize(outputRoot + '.stk', self.normType.get(), self.backRadius.get())          
                                
-            if self.downsampleType.get() == OTHER:
+            if self.doDownsample:
```
delarosatrevin: The coordinates should change if the final samplingRate differs from the input sampligRate. 
This  change can be either because of downsampling or because using other micrographs
which sampling may different from the picking one. So we need to take that into account
for the re-scaling of the coordinates.

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
```diff
@@ -277,76 +268,86 @@ def writePosFilesStep(self):
...
-        fnPosFile =  self._getExtraPath(baseMicName + ".pos")
+        fnPosFile = self._getExtraPath(baseMicName + ".pos")
 
         # If it has coordinates extract the particles      
         particlesMd = 'particles@%s' % fnPosFile
-        
         boxSize = self.boxSize.get()
+        # get box size for extraction, before downsampling
+        if self.doDownsample:
+            boxSize *= self.downFactor.get()
```
delarosatrevin: We should respect the user-selected boxSize, re-scaling should be considered by the wizard.

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
```diff
@@ -277,76 +268,86 @@ def writePosFilesStep(self):
...
-            self.warning(" The micrograph %s hasn't coordinate file! Maybe you picked over a subset of micrographs" % baseMicName)
-                
+            self.warning("The micrograph %s hasn't coordinate file! " % baseMicName)
+            self.warning("Maybe you picked over a subset of micrographs")
+
+        # Let's clean the temporary mrc micrographs
+        pwutils.cleanPattern(self._getTmpPath(baseMicName) + '*')
+
+    def runDownsample(self, stack, boxSize, downFactor):
+        program = "xmipp_image_resize"
```
delarosatrevin: * Why now using xmipp_image_resize instead of xmipp_transform_downsample?
I'm not sure if there is any difference, but just in case. 
* Moreover, why now the downsampling is inside the extractParticlesStep?

delarosatrevin:
@azazellochg , I made some comments through the code. 
We will thinking about moving 'Downsampling factor' parameter next to 'Box size', since it would affect the wizard behavior for the box size. Even, we could remove the question and just ignore when downFactor=1

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
azazellochg: Output coordinates are changed [here](https://github.com/I2PC/scipion/blob/az-extr-particles-dev/pyworkflow/em/packages/xmipp3/protocol_extract_particles.py#L421)

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
azazellochg: - xmipp_transform_downsample is for micrographs only, no? Anyway xmipp_image_resize is the one used in xmipp3 - crop/resize particles protocol
- it has to be after extraction and before normalization and both are in extractParticlesStep

pyworkflow/em/packages/xmipp3/protocol_extract_particles.py
azazellochg: you are right, I will fix this if find some time.

azazellochg:
@delarosatrevin , me too :)
Last time I checked wizard was working.. Downsampling factor is not shown unless you do downsample. I didn't get what question did you mean?

delarosatrevin:
Hi @azazellochg , I getting back to this branch to merge the proposed changes for extract...I plan to include other modification for extraction as well here. But I'm having a failed test, specifically, this one:

./scipion test tests.em.protocols.test_protocols_xmipp_mics.TestXmippExtractParticles.testExtractSort

I think it is because the coordinate position is different from the one associated to the output particle.

azazellochg:
@delarosatrevin ok, I am looking into it now.

azazellochg:
@delarosatrevin , that's strange. if sorting is switched off, the test passes. Otherwise, the objId of particle doesn't match the coords id.

delarosatrevin:
Ok...I will look at it

azazellochg:
ok, it looks like wrong coords assignment:
de8

delarosatrevin:
I finally had the time to go over this refactoring. Following some changes:
* I restored the use of xmipp_transform_downsample program. The idea is, if using downsample, to downsample the whole micrograph before extracting the particles (as was in the devel branch). Moreover, I moved the downFactor to the first section (since this operation is done before any one in the preprocessing) and it is also related to the wizard for boxsize. I removed the boolean asking, now if left 1.0 means no downsampling. 
* I removed separated steps and passed arguments (and programs) needed to be called from the main extractParticlesStep. Apart from reducing the number steps to almost one per micrograph, it have the following advantages. First, now if you only pick particles over a subset of micrographs, the preprocess operations are executed only for those. Second, if we later 'streamify' the extraction of particles (of course, the picking first), it would be better to have only one step for each new micrograph discovered.
* I cleaned-up the handling of different boxsizes. Before, the coordinates were written to .pos file without any modification. Then the xmipp_micrograph_scissor program received a --downsampling parameter and later the coordinates and the metadata where scaled accordingly. Now the coordinates are scaled at the beginning and there is not need to other scaling, apart from the output coordinates. 
* Finally, I partially re-factored the code to follow more our standards. 

@azazellochg , I have run all tests and done other manually. Anyway, this is very critical protocol, could you do some extra tests? Please do not merge the PR yet, since we are having a workshop and we will be using the devel branch, so I should check if we will merge these changes now.

@josuegbl , Could you also do some other tests from your side?

azazellochg:
@delarosatrevin I have run few tests with different coords/mics and both extract protocols (std and for pairs) work correctly, though I will never understand downsampling mics instead of particles during extraction..

delarosatrevin:
Sometimes the downsampling is done in fourier space. This can introduce
some artifacts at the borders. This effect can affect more if downsampling
particles than downsampling the micrographs and extracting from there.

delarosatrevin:
@azazellochg , By the way, we haven't merge this PR yet waiting after the workshop we are having that week. We are using the devel version and we wanted to avoid more changes before the workshop.

azazellochg:
@delarosatrevin no reason to hurry for a merge :) 
I thought it is almost always done in Fourier space, since in real space the artefacts are more severe and it's much slower. Artefacts in Fourier space are possible only if images are not squared, right? Single particles are mostly in squared boxes, except helical processing; while micrographs more often are not (e.g. from K2).

PR #574, opened 2016-09-01
Eraser size in xmipp

### LINKED ISSUE #345: eraser size in xmipp picking GUI
This is more of a feature request: I could not find a way to erase bigger "areas" with bad particles using the default eraser. Maybe smth similar to e2boxer eraser could be implemented (it is a circle with adjustable radius)? 
Btw, I guess after erasing these areas are available again for autopickers? The last one would be difficult to do for all picking protocols, however maybe there is a way? :)

pconesa:
There is now a eraser circle.
Picker Frames (toolbar mainly) have been also refactored: There can be 3 modes and only attributes of the active mode are visible.

Closes #345


josuegbl:
In general, the eraser works good. However, I have a wish list that maybe is easy to implement:
* The definition of the size should be the same as the boxel size , but if you do zoom in/out, change the number that defines your size (not the actual eraser size).
* The eraser should remove the boxes if it touches the external border of the radius of the selected region, not the center of the box.

pconesa:
Done, ready to check.

PR #576, opened 2016-09-02
Vilas delaunay

juannavascalvente:
New functionality has been added to Delaunay to iterate through all faces of the triangulation. This optimizes the performance of the assignment tilt pair method.


delarosatrevin:
Merge should go to devel

PR #595, opened 2016-09-30
First prototype of the ISPyB monitor

delarosatrevin:
Apart from the ISPyB monitor:

- A variable SCIPION_NOGUI and refactored imports (from tkinter and matplotlib.pyplot) to allow Scipion to work in non-gui environment


delarosatrevin:
@josuegbl , Changes in this PR should no affect any of the existing behaviour, since the SCIPION_NOGUI is option. Could you test it any way and approve the PR?

PR #596, opened 2016-10-03
Devel fixes

josuegbl:
remove .txt files produced by OF protocol.


azazellochg:
a very important fix, by the way! few month back I found out that these log files are not text and eating A LOT of disc space.

pyworkflow/em/packages/xmipp3/protocol_movie_opticalflow.py
```diff
@@ -172,13 +171,15 @@ def _processMovie(self, movie):
...
+            
             self._saveAlignmentPlots(movie)
 
         except Exception as e:
             print ("ERROR: %s failed for movie %s.\n  Exception: %s"
                    % (program, inputFn, e))
-
+        
+        # remove this when .txt tmp files are not produced
+        pwutils.cleanPattern(self._getExtraPath("*.txt"))
```
delarosatrevin: Are you sure they are not deleted before other parallel process could be using it?

PR #620, opened 2016-11-02
Project load performance

pconesa:
Few tweaks to improve project load performance:

- Project with Inverted id where making additional queries.
- Protocol updates (from runs.db) are done only if db timestamp is greater (younger)
than protocol timestamp.
- Do not load from db when only need to redraw (selectallProtocols, toggleLables,.)


pconesa:
Please, test or use this branch for a while before merging.

pconesa:
Josu, I've assigned this to you in order play around for a while and see if I screw it somewhere. It's all around project refresh.

pconesa:
closes #633

Issue #621, opened 2016-11-02
parallel ctf


rmarabini:
scipion_devel
protocol: ctffind

 An average cttfind determination takes around 30 sec. Therefore if you run the protocol with 30 CPUs ideally you should be able to solve a CTF each 1 second. Unfortunately, when run in parallel, the protocol star a new step each 4-5 sec therefore, there is no way to run 30 ctffinds in parallel because the workflow engine only starts a new execution each 4-5 seconds. (In practice there is no way to run more than 17 cttfinds in parallel)

delarosatrevin:
Yes, this is one issue that I have recently found with some "working"
program that just waste cpu cycles for a given amount of seconds. If the
working time is less than those 4~5 secs, then you can not have more than
one. I was testing because your report and also some report from Coss.

azazellochg:
Actually, I have a related question.
ctffind3 and ctffind4.0.x support both mpi and threading. Right now in devel branch ctffind3_mp.exe is never used, NCPUS is never set by a protocol; for ctffind4 OMP_NUM_THREADS=1 always. And mpi is switched off in GUI form.py. ctffind4.1.x supports only mpi, but it's not in devel (yet)

Do I understand correctly that at the moment scipion itself implements threading for ctffind (hence the use of ctffind3.exe and not ctffind3_mp.exe)?

pconesa:
And this one?....can this be closed?

PR #638, opened 2016-11-08
Update protocols

azazellochg:
Large update of several protocols:

1. Gautomatch: reorganize GUI form, add exclusive picking options (+ new test), new viewer to display excluded coords and debug output, possible picking without templates
2. Motioncor2: reorganize a bit GUI form, added multi-version support, update params for the latest version, new viewer to display PSD (optional) + global shifts plot. Now possible to save movie stacks. Frame shifts are now written differently (first frame shift is always 0,0)
3. Unblur: added multi-version support, update params
4. Ctffind4: added multi-version support, update params
5. Magdistortion protocols: update test, fix pixel size param, new protocol to correct distortion for coordinates.
6. Gctf: added multi-version support, add phase shift options.
-------**UPD**-------
7. New protocol to create 3D mask with relion
8. IMAGIC-MSA: add error detection after running imagic batch script, added Imagic metadata to 2D classes (d7f5388)

Still TODO:
- [x] upload test data for magdist
- [x] upload bad test coords for gautomatch (exclusive picking)
- [x] upload new soft versions
- [x] think on cuda libs handling for different GPU programs
- [ ] update ctffind to 4.1.6(?) where the bug with output error code is fixed
- [x] check new movie model, dose params, frame numbering, shifts etc
- [x] check issue #621


azazellochg:
@delarosatrevin I was also thinking about cuda libs handling for different programs. It can happen that some GPU programs (now we have 5 of them and relion2 is coming) might use different CUDA versions/libs. I also remember that someone already asked for this for motioncor2 (Juha?).
I suggest that by default cuda libs should be referenced in LD_LIBRARY_PATH (by user, not scipion) and will be used for any GPU program. However, if user wants to provide specific libs for specific program, he can add PROGRAMNAME_CUDA_LIBS var to the config file (as its now done for motioncorr1) and the certain protocol should check if this var is set; if so, then Scipion will replace (or add to the beginning?) any cuda lines in LD_LIBRARY_PATH. What do you think?

pconesa:
@delarosatrevin , would be nice if we can review this together, before you leave.

delarosatrevin:
@pconesa , I have been already reviewing this PR. Yesterday I was passing the test of involved protocols. I can wait until tomorrow to check it together and merge to devel branch.

pconesa:
Cool! As you @delarosatrevin  wish!

delarosatrevin:
Yes, I think it is ready...but it is also a big commit with changes in many protocols. It would be a good opportunity to talk about them.

PR #655, opened 2016-11-16
Refactored ML2D and CL2D

josuegbl:
- createOutputStep was updated.  Now is using classifyItems method instead of readSetOfClasses method.
- Viewers were updated too, taking into account  classifyItems method to generate intermediates classes.
- Several refactors in the protocols and viewers


pyworkflow/em/packages/xmipp3/viewer_cl2d.py
```diff
@@ -81,155 +84,115 @@ def _defineParams(self, form):
...
-                    
-                    for block, fn in files:                        
-                        views.append(self._viewClasses(block, fn, viewParams))
-        else:
-            errors.append('Classes %s have not been produced.' % fnSubset.replace("_", " "))
-        
-        self.errorList(errors, views)
-            
-        return views        
+    # def _setClassOrder(self, classItem, classRow):
```
josuegbl: added a comment: this code is for revision with @cossorzano

PR #659, opened 2016-11-16
Jv multireference alignability

1aviervargas:
CTF correction and downsampling is done inside the protocol to make the use more easy to users. The test has been updated


delarosatrevin:
Merge should be done to devel

Issue #446, opened 2016-05-26
Consider some updates in the Protocols Menu


delarosatrevin:
In the left panel with the protocols menu, some changes we might consider:
- Include new protocols coming from the Scipion-Box branch (several more alignments), maybe create a section for Movie Alignment...and add Monitor under Tools (group 'calculators' under a sub-menu)
- Maybe create a sub-menu only for tools of Scipion-Box
- Maybe have a view of All Protocols? (Maybe automatically generated? Sort alphabetically?
- Change order of Classify and Refine under 3D section

pconesa:
I've added all missing protocols to a "new" section.  We should go through it and decide on placement withing SPA tree  view or whatever.

PR #661, opened 2016-11-16
Update protocols tree

pconesa:
A new "View" with all protocols is added. Nothing static, so it will discover new protocols in the future.


pconesa:
Closes #446 
Closes #661

PR #664, opened 2016-11-21
3D operate

rmarabini:
operate_3D was not working for operations with id > OP_ROW. I fix that and made the code more consistent. That is once the OP_XXXX "macros" are defined everything is defined in terms of these macros, not independently as before. I also added three extra tests.


pconesa:
I like your exploring capabilities, ;-)

PR #668, opened 2016-11-22
Zoom

### LINKED ISSUE #644: implement zoom in graph view
We might need to decide what to do with the text.

pconesa:
closes #644


delarosatrevin:
Great @pconesa!

I have checkout this branch and I was starting to play, but I have found that now Tree Boxes are bigger. I don't know if this is intentional or not. See below (uppper is 'devel', lower is 'zoom' branch)

image

delarosatrevin:
I have been playing and zoom would help a lot! Still I see the boxes bigger at different zoom level, maybe we could also give more space between nodes in the tree. Another minor things that when zooming in, it does not preserve the point where you are zooming. More tomorrow.

pconesa:
I see, thanks! This might be the default font size. I set it up to 10. I'll try with 9. And maybe fot type is different?

PR #690, opened 2016-11-30
fix output path of images.xmd when run in F mode

### LINKED ISSUE #689: ml2d, in F mode fails at convert outputStep
It's a wrapper issue, not Xmipp.

When running it in mlf2d mode, protocol still looks for ml2d_output.....:

00075:   XmippError: FileName::isMetaData: File: 'Runs/000417_XmippProtML2D/ml2d_images.xmd' does not exist
00076:   Protocol failed: FileName::isMetaData: File: 'Runs/000417_XmippProtML2D/ml2d_images.xmd' does not exist

pconesa:
closes #689


pconesa:
ml2d test passes and TestXmippWorflow too.

PR #723, opened 2016-12-20
update local config files with template if requested.

### LINKED ISSUE #267: Error in scipion.conf if there is a previous installation
If there is a previous installation of scipion, when trying to execute ./scipion config there is an error. The output provided is:

Section PACKAGES exists in the template but not in the configuration file.

It is necessary to remove scipion.conf file in the home folder and then it works. It does not work either when executing ./scipion config --overwrite.

pconesa:
fixes #267


azazellochg:
@pconesa while you are looking at it, maybe it is easy to implement alphabetical sorting of lines in the PACKAGES section? In template they are ok, but in output user config file it's a mess. I don't know if ConfigParser() can do this.

pconesa:
I'll have a look

PR #734, opened 2016-12-27
MonoRes fixes

Vilax:
Next changes have been performed

The use of premasked volumes is added.
The gaussian approximation and significance have been removed
Local filter has been implemented as normal mode instead of advance
The region growing has been removed


pconesa:
@Vilax , test was failing. I'm fixing it....

It would be nice to achieve this! What do you think?

image

It can be done though the color key utility, but haven't seen yet if we can do it form the script file.

Vilax:
The key color, or the color bar have to be inserted manually, I tried to find a comand, but, I couldnt. I think it is a manual task.

Now it passes the test. the problem was in the createmask.

pconesa:
I see, do we now already extreme values (min and max) of the resolution?

pconesa:
image
I'm going to look at a way to keep same color key between chimera and slices

PR #776, opened 2017-01-20
Gpu course

rmarabini:
This pull request is not for Scipion 1.1

The main purpose is to provide a framework for GPU programming in xmipp. So far we (coss and myself) have:

1)  created a trivial and useless program called gpu_example that uses GPUs and xmipp metadata.
2) modify scons so gpu compilation is transparent

TODO: so far we have created the infrastructure needed to compile a single program. We need to extend it for libraries. 

NOtE: PATH  should contain information related to the CUDA toolkit, LD_LIBRARY_PATH is not needed unless you plan to execute xmipp_programs outside scipion:

export PATH=$PATH:/usr/local/cuda/bin/
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64


pconesa:
Wow! Something is moving here!

PR #784, opened 2017-01-25
Fix validation error

kevinpetersavage:
This is a minor change to fix a case where alignment validation throws an exception instead of registering an error. 

Can this be merged into devel?

### DIFF
--- a/pyworkflow/em/packages/grigoriefflab/protocol_summovie.py
+++ b/pyworkflow/em/packages/grigoriefflab/protocol_summovie.py
@@ -29,7 +29,7 @@ from os.path import exists, realpath, abspath
 import pyworkflow.protocol.params as params
 import pyworkflow.protocol.constants as cons
 import pyworkflow.utils.path as pwutils
-from pyworkflow.em.protocol import ProtAlignMovies
+from pyworkflow.em.protocol import ProtAlignMovies, ProtProcessMovies
 
 from grigoriefflab import SUMMOVIE_PATH
 from convert import writeShiftsMovieAlignment
@@ -134,7 +134,12 @@ class ProtSummovie(ProtAlignMovies):
         return []
 
     def _validate(self):
+
+        if self.inputMovies.get() is None:
+            return
+
         errors = []
+
         inputMovies = self.inputMovies.get()
         firstFrame, lastFrame, _ = inputMovies.getFramesRange()
         frames = lastFrame - firstFrame + 1
--- a/pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
+++ b/pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
@@ -186,18 +186,20 @@ class ProtUnblur(ProtAlignMovies):
         return []
 
     def _validate(self):
+
         errors = []
         if not os.path.exists(UNBLUR_PATH):
             errors.append(
                 "Cannot find the Unblur program at: " + UNBLUR_PATH)
+        if self.inputMovies.get():
 
-        if self.doApplyDoseFilter:
-            inputMovies = self.inputMovies.get()
-            doseFrame = inputMovies.getAcquisition().getDosePerFrame()
+            if self.doApplyDoseFilter:
+                inputMovies = self.inputMovies.get()
+                doseFrame = inputMovies.getAcquisition().getDosePerFrame()
 
-            if doseFrame == 0.0 or doseFrame is None:
-                errors.append('Dose per frame for input movies is 0 or not '
-                              'set. You cannot apply dose filter.')
+                if doseFrame == 0.0 or doseFrame is None:
+                    errors.append('Dose per frame for input movies is 0 or not '
+                                  'set. You cannot apply dose filter.')
 
         return errors
 
--- a/pyworkflow/em/packages/xmipp3/protocol_movie_opticalflow.py
+++ b/pyworkflow/em/packages/xmipp3/protocol_movie_opticalflow.py
@@ -234,24 +234,28 @@ class XmippProtOFAlignment(ProtAlignMovies):
     #--------------------------- INFO functions -------------------------------
     def _validate(self):
         errors = ProtAlignMovies._validate(self)
-        # Although getFirstItem is not remonended in general, here it is
+        # Although getFirstItem is not recomended in general, here it is
         # used olny once, for validation purposes, so performance
         # problems not should be apprear.
-        inputSet = self.inputMovies.get()
-        movie = inputSet.getFirstItem()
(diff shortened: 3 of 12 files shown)


pconesa:
Thanks @kevinpetersavage, actually I'm realising that some methods raise the same message twice (when motioncor) and others still fails because are not using its parent (ProtAlignMovies) validation method. @delarosatrevin , since you seems to be around.....does it make sense to make all "childrens" (inheritance) of ProtAlignMovies invoke ProtAlignMovies_validate first (motioncorr does it).

pconesa:
Umm! summovie does not call super validate for a good reason, it does not define, at least cropDimX, which is used in the ProtAlignMovies._validate()

pconesa:
And motioncorr2, if  "doApplyDoseFilter" fails to...Thanks @kevinpetersavage, this seems a "gold mine".

pconesa:
@kevinpetersavage , I've made some changes to contemplate validation exceptions, in case they happen. Could you please tell me which "movie alignment protocol" validation you where trying to fix?

kevinpetersavage:
@pconesa it was motincor2 so you may already have found it! Thanks for looking into this.

delarosatrevin:
@pconesa , Maybe we should review all classes that inherits from that base protocol. In principle the validations in the base class should be common for all derived classes. In some exceptional cases, some protocols do not define all the base properties, which can lead to exceptions...but those should be detected while developing the protocol.

pyworkflow/em/packages/grigoriefflab/protocol_summovie.py
```diff
@@ -134,7 +134,12 @@ def _methods(self):
         return []
 
     def _validate(self):
+
+        if self.inputMovies.get() is None:
+            return
```
delarosatrevin: @pconesa, I think that empty movies selection is an error that should be notified to the user. It should be something like:

return ['Select input movies']

Maybe we should look at the same error in other protocols and be consistent with the message.

pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
```diff
@@ -186,18 +186,20 @@ def _methods(self):
         return []
 
     def _validate(self):
+
         errors = []
         if not os.path.exists(UNBLUR_PATH):
             errors.append(
                 "Cannot find the Unblur program at: " + UNBLUR_PATH)
+        if self.inputMovies.get():
```
delarosatrevin: Report the error or not, depending if it is validated in the base class.

pyworkflow/em/packages/xmipp3/protocol_movie_opticalflow.py
```diff
@@ -234,24 +234,28 @@ def _processMovie(self, movie):
     #--------------------------- INFO functions -------------------------------
     def _validate(self):
         errors = ProtAlignMovies._validate(self)
-        # Although getFirstItem is not remonended in general, here it is
+        # Although getFirstItem is not recomended in general, here it is
```
delarosatrevin: recommended

pyworkflow/em/protocol/protocol_align_movies.py
```diff
@@ -243,8 +243,8 @@ def _validate(self):
         # self.inputMovies.get().close()
         # frames = movie.getNumberOfFrames()
 
-	if self.inputMovies.get() is None:
-            errors.append("Input movies is None")
```
delarosatrevin: I think we should report the error even if we do not proceed with other validations

pyworkflow/em/protocol/protocol_import/coordinates.py
```diff
@@ -94,8 +97,9 @@ def createOutputStep(self, importFrom, *args):
         coordsSet = self._createSetOfCoordinates(inputMics)
         coordsSet.setBoxSize(self.boxSize.get())
         ci = self.getImportClass()
-        
+        print "HELLO"
```
delarosatrevin: Remove debug print

pyworkflow/em/protocol/protocol_import/coordinates.py
```diff
@@ -94,8 +97,9 @@ def createOutputStep(self, importFrom, *args):
...
         for coordFile, fileId in self.iterFiles():
+            print "Coord and fileId", coordFile, fileId
```
delarosatrevin: More debug code

pyworkflow/protocol/protocol.py
```diff
@@ -1397,9 +1397,15 @@ def validate(self):
...
             errors += ['*%s* %s' % (label, err) for err in paramErrors]                
         # Validate specific for the subclass 
-        childErrors = self._validate()        
-        if childErrors:
-            errors += childErrors
+        try:
+            childErrors = self._validate()
+            if childErrors:
+                errors += childErrors
+        except Exception as e:
```
delarosatrevin: Exception in the validation methods are only programming errors, not users errors, so we should be clear in differentiating them.

pyworkflow/em/packages/grigoriefflab/protocol_summovie.py
pconesa: Actually, form takes care of this validation, since inputMovie is a pointer. User is warned about it.

pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
pconesa: Same, as above.

pyworkflow/em/packages/xmipp3/protocol_movie_opticalflow.py
pconesa: ups!

pyworkflow/em/protocol/protocol_align_movies.py
pconesa: Again, I tested all movie alignment protocols with empty movie and you get the warning.

pyworkflow/em/protocol/protocol_import/coordinates.py
pconesa: uups!

pyworkflow/protocol/protocol.py
pconesa: I agree. Any ideas, An apology? Sorry....this is embarrassing...

pyworkflow/em/packages/grigoriefflab/protocol_summovie.py
pconesa: Now all movie alignment protocols warn this way, when input movies is empty:
image

PR #801, opened 2017-02-09
tolerate and infrom error at protocol update check

pconesa:


pyworkflow/protocol/protocol.py
```diff
@@ -1737,9 +1737,14 @@ def isProtocolUpToDate(protocol):
 
     dbTS = getFileLastModificationDate(protocol.getDbPath())
 
-    if protTS < dbTS:
-        return False
-    else:
-        return True
+    try:
+
+        if protTS < dbTS:
```
delarosatrevin: This return can be written in a more compact way:

return protTS >= dbTS

pyworkflow/protocol/protocol.py
```diff
@@ -1737,9 +1737,14 @@ def isProtocolUpToDate(protocol):
 
     dbTS = getFileLastModificationDate(protocol.getDbPath())
 
-    if protTS < dbTS:
-        return False
-    else:
-        return True
+    try:
```
delarosatrevin: A try/catch seems too much for this validation, it could be something like:
if not (protTS and dbTS):
    print "Error message"....
else:
   .....

pconesa: yep, make sense. I'll improve it on monday

pyworkflow/protocol/protocol.py
pconesa: ok

PR #822, opened 2017-02-15
select up, highlight neigbourghs

pconesa:
and age color mode rescue.


pconesa:
@josuegbl could you play around with this and feedback.

Issue #833, opened 2017-02-20
allow silent config


pconesa:
now you must enter the notify!! value....for buildbot and other programmatic installation this is a problem

delarosatrevin:
Yeap, good point @pconesa . Maybe we should add a parameter such as: --notify [no/yes] for avoiding this question to pop up in automatic installations.

rmarabini:
done

PR #839, opened 2017-02-22
Rm package statistics. skip user intervention

rmarabini:
incorporate --report flag. Then main goal is to allow automatic installations since now an answer from a user is required in the configuration process.

   if --report enables SCIPION_REPORT is set to True without asking to the user.

      ./scipion config --report

This pull request is an  answer to the issue #833 opened by @pconesa


delarosatrevin:
@rmarabini, The variable name is SCIPION_NOTIFY, not SCIPION_REPORT...can you change the parameter name (--notify instead of --report) to have a more consistent naming?

rmarabini:
done, refactored so "report" is now "notify"

PR #847, opened 2017-02-23
Compilation with GPU

cossorzano:


config/templates/scipion.template
```diff
@@ -90,6 +90,11 @@ NVCC_LIBDIR = /usr/local/cuda/lib64
 
 MATLAB_DIR = /usr/local/MATLAB/R2011a
 
+# nVidia Compiler
```
pconesa: Even there's a similar file at: https://github.com/I2PC/scipion/blob/devel/software/em/xmipp/applications/programs/gpu_example/gpu_example_main.cpp

PR #871, opened 2017-03-01
Rm fsc button

rmarabini:
Now the createFSC button works as expected (the button appears in FSC plot create from "in-memory" FSC). I have modified relion-refine and relion-postprocess so they use it.  

NOTE: I see that I have branched from rm_scipionbox (I previous pull request)


pconesa:
@rmarabini , and you keep requesting a pull to release-1.1 fixes...Please for this, branch from release-1.1 and PR ro release-1.1 if apply.

PR #875, opened 2017-03-02
relax expected value for phantom project case 5

pconesa:


pconesa:
hi @cossorzano , can you check if this sound correct or is there a need to look into why the value has increased for the 5th case of the phantom tests

PR #888, opened 2017-03-08
Co jiri

pconesa:
Optimizations


cossorzano:
This is my pull request, so it is better if Roberto verifies

delarosatrevin:
@rmarabini, Can you take a look to this PR? It seems to be about some changes in Xmipp internal code.

@pconesa I would propose that if @rmarabini can not check this...then that @cossorzano double-check the changes does not affect the expected behavior and merge the PR himself.

PR #897, opened 2017-03-10
Merge motioncorr package.

pconesa:
both motioncorr packages merged with different cuda binaries inside


pconesa:
@delarosatrevin , this PR is blocked by travis. I'm looking into travis...and not sure if is actually doing anything useful. Am I wrong?

delarosatrevin:
@pconesa I has happened before that sometimes Travis keeps in yellow the PR blocked. Travis is a continuous integration tools that can be linked with a GitHub repository. We are only using it to enforce the rule of PR, i.e. that code can not be commited to protected branches directly.

pconesa:
I see! I thought that github has it's own way of protecting a branch.

PR #898, opened 2017-03-13
updated extract particle protocol

josuegbl:


pconesa:
Something is not working @josegutab. Extract particle in TesXmippWorkflow dos not fail, but it does not generate any set of particles:
image

delarosatrevin:
Maybe we can add to the test a simple check that the output is not None.

pconesa:
Well, it fails right after when the next protocol try to use the output. 
I mean the test actually fails but in the next "box"

pconesa:
related to #905

pconesa:
should fix #903

PR #906, opened 2017-03-14
Add new protocol ProtExportEMDB

ajimoreno:
A new protocol (ProtExportEMDB) to generate the volume and FSC files using the format described in EMDB.


pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+import os
+
+
+class ProtExportEMDB(EMProtocol):
+    """ stress  will  stress  test  a  computer system in various selectable
+       ways. Several options require the program stress-ng.
+    """
+    _label = 'exportEMDB'
+    _program = "" 
+    _version = VERSION_1_2
```
delarosatrevin: @pconesa I still do not like the _version usage only for the "new" state in the GUI. I find it very confusing. It should be named like _scipion_previous_stable_when_protocol_appeared. Just _version could mean protocol version? scipion version when the protocol apeared/updated? (neither of these cases)

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+from pyworkflow import VERSION_1_2
+from pyworkflow.em.convert import ImageHandler
+import os
+
+
+class ProtExportEMDB(EMProtocol):
+    """ stress  will  stress  test  a  computer system in various selectable
+       ways. Several options require the program stress-ng.
+    """
+    _label = 'exportEMDB'
```
delarosatrevin: @ajimoreno This label does not follows the convention of most other protocol. It should be all lowercase, maybe something like 'export emdb'. Even for PDB we have use lowercase in other cases.

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
pconesa: what about something like a date? This was my first attempt, but could't see a way to compare a date with a release...unless we have a release date somewhere. The benefit of this approach would be that "new" could disappear after a while, even within the same release.

I agree var name is confusing....I can try to think of a better name if that's enough.

delarosatrevin: Yep, a data could be a good idea. Something like:
_relasedate = '2017/03/14'

We already have that for the released version here: 
https://github.com/I2PC/scipion/blob/master/scipion#L43

pconesa: I've added an issue to refactor this.

delarosatrevin: Thanks @pconesa!

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+        message = "Data Available at directory: *%s*"% os.path.abspath(self._getExtraPath())
+
+        return [message]
+
+    def _methods(self):
+        return []
+
+#--------------------------- UTILS functions ---------------------------------------------------
+
+    def getName(self):
```
josuegbl: There are some path funtions in pyworkflow.utils.path that could be help here to obtain the output path files.

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+    """ stress  will  stress  test  a  computer system in various selectable
+       ways. Several options require the program stress-ng.
+    """
+    _label = 'exportEMDB'
+    _program = "" 
+    _version = VERSION_1_2
+
+    def __init__(self, **kwargs):
+        EMProtocol.__init__(self, **kwargs)
+        self.xmippMic={}
```
josuegbl: Which is the use of this dictionary? in any case, xmippMic is an unlucky name since this protocol does not belong to xmipp.

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+    def _defineParams(self, form):
+        form.addSection(label='Input')
+
+        form.addParam('exportVolume', params.PointerParam, label="Volume to export", important=True,
+                      pointerClass='Volume',
+                      help='This volume will be exported using mrc format')
+        form.addParam('exportFSC', params.PointerParam, label="FSC to export", important=True,
+                      pointerClass='FSC',
+                      help='This FSC will be exported using XML emdb format')
+
```
josuegbl: It may be possible to add a new parameter with a desired path to store the volume and the FSC? Maybe it is useful for the users if they had the capacity to choose a directory to store the files.

pyworkflow/em/protocol/protocol_export/protocol_export_emdb.py
```diff
@@ -0,0 +1,114 @@
...
+    #--------------------------- INSERT steps functions --------------------------------------------
+    def _insertAllSteps(self):
+            self._insertFunctionStep('exportVolumeStep')
+            self._insertFunctionStep('exportFSCStep')
+
+    #--------------------------- STEPS functions --------------------------------------------
+
+    def exportVolumeStep(self):
+
+        self.nombreVolumen, self.nombreFsc = self.getName()
```
josuegbl: variable names should be in english

PR #911, opened 2017-03-17
add latest motioncor2

azazellochg:
closes #797 
To summarize:
1. Correction of anisotropic magnification - added input params, also Grigorieff's mag dist estimate protocol can be used as input
2. Set threads =1 since it doesn't use CPUs.
3. Fix errors when only DW sum was saved

PR #920, opened 2017-03-21
Stream continue stop

delarosatrevin:
### This PR fixes issues mainly related with streaming:
- Now protocols can be continued (skip done movies, created stacks and computed ctf)
- Now the import in streaming can be explicitly stopped by the user (right click - STOP STREAMING)
- Now frames can be stacked even if not using streaming.
- Added test to launch an streaming session and another to test creating stack from single frames.
- Thumbnails of the micrographs can be computing using e2proc2d.py to improve further visualization.

### Some by-products (or side effects):
- Now protocols can have special actions in the GUI (if they implement special function getActions), used to stop the streaming
- Implemented in the mapper the function maxId() to solve a problem with the last inserted id in the Set class


delarosatrevin:
I have launched some related tests...and I'm now launching a complete streaming session (transferring a real data from other machine) and it is going well. Anyway some additional tests would be good.

delarosatrevin:
I just suffered a failure from ctffind (input mic was missing)...I added support to continue, with the same approach as movies, i.e. reading the done.txt file.

delarosatrevin:
Merged from #911

josuegbl:
@delarosatrevin, @azazellochg, I have a question about the last verson of motioncor2. All the new parameters (related to magnification correction) added, belongs also to the previous versions of motioncor2, or its a new implementation in the new version? If so, I would recommend that this new params only appears if you have installed this motioncor2's version.

delarosatrevin:
I think that there are only available in the new version of motioncor2...nonetheless, I disagree with you. I think we should show them (in the other case the user can miss the point the we have updated the protocol to support the newer version) and then the _validate method can warn if you do not have the proper version to use that functionality. By the way, I haven't tested that part of the new parameters.

josuegbl:
I think there are other ways to show if we support the latest (or certain) versions of the programs; for instance with scipion install --help. Maybe we need to think a more cleaver way, but if your version does not support certain params and you can not install the last version, because you are not the system manager or your system can not support the version, could be confuse for the users this additional parameters, that is only supported if you have the latest version, or the version from which its supported. Consider also that you need to install CUDA-8.0.
This case is the same for relion 2.0, that we support but if you have not installed, the GPU related parameters not appears. 
Maybe we need establish a consensus for this cases to gain in uniformity

delarosatrevin:
Ok, you also have some good arguments. My concern is only to make really visible to the users that we support this version as well...and then they should complain to the sysadmin to installed if not there.

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
```diff
@@ -140,34 +140,49 @@ def _defineAlignmentParams(self, form):
...
-                       help='Major scale factor.')
-        group.addParam('scaleMin', params.FloatParam, default=1.0,
-                       condition='not useEst and useMotioncor2 and doMagCor',
-                       label='Minor scale factor',
-                       help='Minor scale factor.')
-        group.addParam('angDist', params.FloatParam, default=0.0,
-                       condition='not useEst and useMotioncor2 and doMagCor',
-                       label='Distortion angle (deg)',
-                       help='Distortion angle, in degrees.')
+        if getVersion('MOTIONCOR2') == '01302017':
```
delarosatrevin: What would happens when comes a new version of MC2 that also support this options?

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
```diff
@@ -140,34 +140,49 @@ def _defineAlignmentParams(self, form):
...
+                           label='Minor scale factor',
+                           help='Minor scale factor.')
+            group.addParam('angDist', params.FloatParam, default=0.0,
+                           condition='not useEst and useMotioncor2 and doMagCor',
+                           label='Distortion angle (deg)',
+                           help='Distortion angle, in degrees.')
+        else:
+            form.addParam('motioncor2Version', params.LabelParam,
+                          condition='useMotioncor2',
+                          label='It seems motioncor2-v01302017 is not '
```
delarosatrevin: Same here, too specific for the current version

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
azazellochg: if getVersion('MOTIONCOR2') not in ['03162016', '10192016'] would be better, no?

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
josuegbl: I dont know how to handle this. Just change the test for "It seems last version of motioncor2 is not..." is not valid because at some point it will not be the "last version" that we are referring now. Any suggestions?

azazellochg: @josuegbl condition can use any function now, right? you could create a function checking if getVersion('MOTIONCOR2') not in ['03162016', '10192016']

josuegbl: @azazellochg, I going to create the function. Thanks.

delarosatrevin: yes @azazellochg I think this is a good idea. A function like: 'supportsMagCorrection' and internally implement how this is figured out. In our case now check the version in the list. I think anyway that the point of @josuegbl here was about the message. Maybe something like:
"Scipion supports some versions of motioncor2 that can do magnification correction, but they do not seems to be installed. Check available versions with: 'scipion install --help'

PR #925, opened 2017-03-23
Release 1.1 fixes

josuegbl:
We should test the following:
- [x] Installation and execution of new motioncor2. Also the backward compatibility of the protocol. 
- [x] Images smaller than 50 px should be sorted after extraction. A test for this can be useful.
- [x] Test if set of averages (from solid angles protocols) has alignment proj.
- [x] Test the new implementation of extract movie particles protocol.


delarosatrevin:
@josuegbl , By accident I merged the branch related to the new version of motioncor2 here. It was supposed to be tested in the streaming branch here #920 . So I would recommend to focus in the other three points. Anyway, I have updated #920 with this branch and I'm testing the new version of motiocor2 and seems to be working. I have not tested the installation part, since I'm using a local installation of the new version here.

josuegbl:
@delarosatrevin, ok, Im going to follow your recommendation, although motioncor2 must be tested before merging.

delarosatrevin:
@josuegbl Sure, you can test the other points directly in #920 (since I have merged this branch to there)...and then proceed with motioncor2 (It has been working successfully in my hands)

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
```diff
@@ -131,6 +132,35 @@ def _defineAlignmentParams(self, form):
         form.addParam('tol', params.FloatParam, default='0.5',
                       label='Tolerance (px)', condition='useMotioncor2',
                       help='Tolerance for iterative alignment, default *0.5px*.')
+
```
pconesa: @azazellochg , in Relion we have a similar case, and @josuegbl is adding the params based on the version installed, so user is not presented with parameters he can't use. To inform the user about the existence of a newer version we can  add a label in the form like "There is a new version of motioncor2....". 
Also yesterday we agreed with @delarosatrevin that we need so decouple the MOTIONCORR_CUDA_LIB and have  new MOTIONCOR2_CUDA_LIB, so both morioncorX packages can be used simultaneously.

@azazellochg, we can do this if you are busy! Just let us know.

Thanks again.

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
```diff
@@ -160,7 +192,7 @@ def _defineAlignmentParams(self, form):
                            "dose-dependent filter to the frames")
 
         # Since only runs on GPU, do not allow neither threads nor mpi
-        form.addParallelSection(threads=3, mpi=1)
+        form.addParallelSection(threads=1, mpi=1)
```
pconesa: I might be missing something here, but why not allowing more threads?, We did some tests here and our results were that best performance is achieved with 2 "threads" in the GPU, which corresponds to 3 threads (i know this sounds counterintuitive):
image

delarosatrevin:
@pconesa I agree with @azazellochg that the default option should be one, since it seems to me the safer one. I think the measurements of performance are very dependent on the input parameters, like the size of the input movies or the number of frames, or even the type of gpu card that you have. How many different input combinations did you use for the performance tests?

pconesa:
I didn't do a bwide test of the parameters, only what you saw. Then 
leave it as it is. It's fine.

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
azazellochg: @pconesa alright, you convinced me :) You could do this since I'm a bit busy next 1-2 weeks

pyworkflow/em/packages/motioncorr/protocol_motioncorr.py
azazellochg: @pconesa hm, I do not understand, how does threads option work? If it determines number of GPUs and not CPUs, why do we need GPU ID field at all?

pconesa:
Threads or MPI will be  created by Scipion to send as many "movie 
alignment" processes to the GPU/s

The GPU ids will be sent to motioncor2 and is motioncor2 who is 
splitting the SAME "movie alignment" between the GPUs.

Lets say you set threads = 1 GPU=0

This will be a sequential process:  1 movie at a time on GPU 0

If you want 1 thread and 2 GPUs (0,1):

This will be split by motioncor2 between the 2 GPUs but the same movie 
is process, but it still can be seen as a "sequential alignment" process.

If you want 3 threads (2 is not an option, cause one thread will be use 
for the master process) and 1 GPU (0)

Then two movies are sent to GPU 0, so at the same time two movies are 
being aligned.

At least this is how I understood it.

delarosatrevin:
@azazellochg The parallelization is done in the way Pablo has explained. In some cases, the motioncor2 does not saturate the gpu memory or usage, so you can allocate two (or sometimes even more) processes per gpu card. That's is where the Threads parameter in the Scipion protocol comes to help...but as you known the motioncor2 does not support threading internally (at least now).

azazellochg:
@pconesa @delarosatrevin thank you both! Made things clear for me.

Issue #930, opened 2017-03-27
fix notifier.py for scipion old version


azazellochg:
> Dear colleagues
>  we experiencing the  problem going from the old scipion version 1.0.1 to
> the new one 1.1
> 
> The error we are getting trying to open the project is Error 2 No such file
> or directory: 'Logs/data.log'
> 
> In the version 1.0.1. you can open the project and don't have this error
> 
> Sincerely
> Dmitry

pconesa:
Thanks Gregory!

rmarabini:
Summary: when a project created with scipio 1.0 is open with scipion 1.1 the follwing Error appears

   " Error 2 No such file or directory: 'Logs/data.log'"

1) It seems that the notifier system (new in scipion 1.1) creates a file called data.log in the directory $PROJECT_NAME/Logs.
2) it seems that scipion 1.0 does NOT create this directory by default (scipion 1.1 does)
3) notifier routine has been modified so it checks and creates the Log directory (if needed) before writting the data.log file

  I close this issue

pconesa:
Great! many thanks Roberto.

PR #932, opened 2017-03-28
fix issue 'fix notifier.py for scipion old version #930'

rmarabini:
this fix create the Log directory if it does not exists


pyworkflow/gui/project/notifier.py
```diff
@@ -117,6 +117,12 @@ def notifyWorkflow(self):
             #print "No change: Do not send new data"
             return
         else:
+            # For compatibility with version 1.0 check 
+            # if Log directory exists. If it does not 
+            # create it
+            #TODO REMOVE this check in scipion 1.3
+            import distutils.dir_util
```
delarosatrevin: @rmarabini , I think that you should try to use the function we already have and avoid additional imports.
Indeed, you don't need to add an spurious default parameter to just get the directory here name. It could be something like:

dataFile = self._getDataFileName()
import pyworkflow.utils as pwutils # this import should go in the top if not already imported
pwutils.makeFilePath(dataFile) # this creates the folder of the file path if not exists

rmarabini:
changes suggested by #delarosatrevin has been incorporated

pconesa:
Thanks @rmarabini and @delarosatrevin , I'm merging this.

PR #944, opened 2017-04-07
Monores v2

Vilax:
The MonoRes code has been revisited and enhanced in performance.

Some bugs have been solved. They were related with the calculation of the monogenic amplitude


delarosatrevin:
@Vilax  I think you have merge stuff from the devel in your branch that are not supposed to be included in the release version (mostly some GPU stuff). Maybe the easy way is to create a new branch from release-1.1 and just copy the changes related to your protocol...and make the PR from that one.

PR #945, opened 2017-04-07
monores changes for pull request

Vilax:
The performance of monores has been enhanced

A bug in the calculation of the monogenic amplitude has been solved


pconesa:
@Vilax , good work. All test are passing but I've seen a weird visualization behaviour:

Steps:
1.- Load a color slices plot. (do not close it)
2.- Load  the histogram----histogram is painted in the slices plot (color map figure).

image

PR #947, opened 2017-04-07
spider fixes

### LINKED ISSUE #946: spider protocols fix
1. ap sr must support threading
2. change csh to sh, since there are csh versions that do not work with currect script synthax

azazellochg:
1. threading now works in most protocols (refinement not tested since there is no test protocol)
2. change csh to sh for more compatibility
3. add output generation in spider reconstruct protocol
Closes #946

### DIFF
--- a/pyworkflow/em/packages/spider/scripts/mda/apsr4class.msa
+++ b/pyworkflow/em/packages/spider/scripts/mda/apsr4class.msa
@@ -3,6 +3,7 @@
 ; Using SPIDER command AP SR
 
 ; ---------------- Parameters ----------------
+x30 = 0                ; number of threads to use, 0 = use all
 [inner-rad] = 5        ; first ring radius for alignment, pixels
 [outer-rad] = 44       ; expected object radius, pixels
 ;;[obj-diam] = 88        ; expected object diameter, pixels
@@ -35,9 +36,7 @@ vm
 echo "Performing reference-free alignment"; date
 
 vm
-echo "if(! -d [apsr_dir]) mkdir -p [apsr_dir]"|csh
-;vm
-;echo "if(! -d [aligned_dir]) mkdir -p [aligned_dir]"|csh
+echo "[ ! -d [apsr_dir] ] && mkdir -p [apsr_dir]"|sh
 
 ; GET IMAGE DIMENSION
 
@@ -77,10 +76,10 @@ _2      ; OUTPUT
 (3)     ; Gaussian low-pass
 (0.02)  ; filter radius
 
-;; use all processors
-;md
-;set mp
-;(0)
+; use x30 processors
+md
+set mp
+x30
 
 ; run reference-free alignment
 AP SR
--- a/pyworkflow/em/packages/spider/scripts/mda/apsr4class.msa
+++ b/pyworkflow/em/packages/spider/scripts/mda/apsr4class.msa
@@ -3,9 +3,9 @@
 ; Using SPIDER command AP SR
 
 ; ---------------- Parameters ----------------
-x30 = 0                ; number of threads to use, 0 = use all
 [inner-rad] = 5        ; first ring radius for alignment, pixels
 [outer-rad] = 44       ; expected object radius, pixels
+[nummps]    = 0        ; number of processors to use (0=all)
 ;;[obj-diam] = 88        ; expected object diameter, pixels
 
 ; ---------------- Input files ----------------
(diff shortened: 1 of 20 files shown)


pyworkflow/em/packages/spider/scripts/mda/apsr4class.msa
```diff
@@ -3,6 +3,7 @@
 ; Using SPIDER command AP SR
 
 ; ---------------- Parameters ----------------
+x30 = 0                ; number of threads to use, 0 = use all
```
delarosatrevin: @azazellochg Do you mind using [nummps] here? Just to provide a more clear name and be consistent with the other scripts

azazellochg: @delarosatrevin sure, you can modify that.

delarosatrevin:
I have passed all test related to Spider and seems to be working, except:
1) The test: scipion test tests.em.workflows.test_workflow_spiderMDA.TestSpiderConvert is failing, but this not seems to be related with these changes
2) Even if the MPI=0 in the addParallelSection, it is been display in the form...but this seems like a general issue.

@pconesa Can you check about 2) ? or maybe @ldelcano that changed to 1 or 0 recently.

azazellochg:
@delarosatrevin hm, for me this test works..

pconesa:
Test works for me too. I'll check the MPI thing now.

pconesa:
If I'm right, @ldelcano only reviewed all protocols that has STEPS_PARALLEL and none of the Spider protocols are parallelized this way. We haven't touched the form, so it might be an older issue. What should be the behavior in the form for protocols not parallelized by scipion?:
None in any (Thread or MPI should hide the field?)

delarosatrevin:
It has been a long time since I have modified the form.py core file, but I think if any of threads or mpi is equal to 0 in the definition of the parameters (addParallelSection), they should not appear in the form. Zero means that are not allowed.

pconesa:
As long as the protocol has the attribute it will show up:

`self.allowMpi = hasattr(self, 'numberOfMpi')`

And in addParallelSection only if mpi > 0 it will add the parameter. So you are guessing rith @delarosatrevin . I guess we want no MPI field in these spider protocols, right?

pconesa:
Issue is here...and probably with all protocols that have multiple inheritance, or at least invoking 2 __init__:

`class SpiderProtAlign(ProtAlign2D, SpiderProtocol):
    """ Base class for Spider alignment protocols. """
    
    def __init__(self, script, alignDir, **args):
        ProtAlign2D.__init__(self, **args)
        SpiderProtocol.__init__(self, **args)
`

With the first __init__ call , allowMpi is set correctly to false, but at the same time numberOfMpi is set to 0.
Within the second __init__ the object has the attribute allready numberOfMpi set to 0 and therefore allowMpi is set to True.

pconesa:
@delarosatrevin this will fix the issue for multiple inheritance protocols:

`       # NOTE: For multiple inheritance protocols code is passing here twice
        # and second time numberOfMpi attribute is set to 1.
        if not hasattr(self, 'allowMpi'):
            self.allowMpi = hasattr(self, 'numberOfMpi')`

I feel is a bit of a hack....but behaves and seems safe enough. Any cleaner fix?

Same for threads.

delarosatrevin:
I'm not sure...what about if the second pass is the proper one? This will make us to review now all multiple inheritance. Anyway I think the way it is now is not good.

pconesa:
Yep, you are right, I realized that while driving back.

We'll review how many multiple inheritance there are and will see.

May be for this we can set the allowMpi to false after the second init 
and review it later with the rest.

What do you think?

delarosatrevin:
Yep, I will here focus in Spider protocols (for most of which I think they do not support MPI) and then think in a more general solution for the future.

azazellochg:
@delarosatrevin indeed, very few support mpi natively: https://spider.wadsworth.org/spider_doc/spider/docs/mpi.html

pconesa:
Thanks @azazellochg and @delarosatrevin , looking at your the link.....I see that we need to use "spider_linux_mpi_opt64" and this is not found in our code. Can I assume that none of our Spider protocols are currently implemented to make use of the MPI capabilities of Spider'

azazellochg:
@pconesa scpider-reconstruct and spider-refinement could use mpi, but it's not implemented for now.

pconesa:
Thanks Gregory! I think MPI functionality will not go in the 
release....unless any of you insist and see it very, very clear and safe.

How do you see it? Is it obvious?

delarosatrevin:
I would say that it is not important for now.

azazellochg:
i would leave them "as is" for now, since they are not even fully finished (using dummy zero angles etc)

pconesa:
So, all of the Spider protocols show now only "Threads".

PR #959, opened 2017-04-18
Hide disabled protocols (to hide relion picking protocols)

pconesa:
Added  isDisabled class method for protocols.
Test isDisabled when loading protocols, searching, of when loading the protocols.conf. Disable relion picking protocols dynamically, based on installed version


delarosatrevin:
@pconesa I like this workaround...I will test it later and merge the changes.

pconesa:
Ok, don't worry, if you like it Josue will test it. Thanks Jose.

PR #969, opened 2017-04-26
Release 1.1 fixes

pconesa:
Initial Project window size and layout tweaks of frames.


pyworkflow/gui/project/project.py
```diff
@@ -115,7 +115,10 @@ def __init__(self, path, master=None):
         self.showGraph = False
         Plotter.setBackend('TkAgg')
         ProjectBaseWindow.__init__(self, projTitle, master,
-                                   icon=self.icon, minsize=(900,500))
+                                   icon=self.icon, minsize=(90,50))
```
delarosatrevin: Why minsize has been set so small?

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -602,15 +603,14 @@ def createContent(self):
...
-
         bgColor = Color.LIGHT_GREY_COLOR
         # Left pane, contains Protocols Pane
         leftFrame = tk.Frame(p, bg=bgColor)
         leftFrame.columnconfigure(0, weight=1)
         leftFrame.rowconfigure(1, weight=1)
 
         # Protocols Tree Pane
-        protFrame = tk.Frame(leftFrame, width=300, height=500, bg=bgColor)
+        protFrame = tk.Frame(leftFrame, width=30, height=20, bg=bgColor)
```
delarosatrevin: Same here, why so small?

pyworkflow/gui/project/project.py
pconesa: Javi is having trouble in a cluster with a really small resolution login node...and obviously 90,50 is not operative at all but only the user will dimension it to 90,50. Actually I just removed a 0. Didn't consider important the minsize.

pyworkflow/gui/project/viewprotocols.py
pconesa: This was some some test and for some reason, it does not follow this values. It does'nt matter the values there. I'll set old values back.

pconesa:
@delarosatrevin ... add the fixes...It could be now or "never". ;-)

rmarabini:
cloased

PR #992, opened 2017-05-08
New protocol_ctf_selection added

ajimoreno:
A protocol to make a selection of meaningful CTFs in basis of the defocus values, the astigmatism, and the resolution is added. This protocol is designed to work in streaming mode.
A test to verify the correct performance of the previous protocol is also added, "test_protocols_ctf_selection".


josuegbl:
A few comments about this pull request:
- This protocol does not uses any xmipp programs, so the name XmippProtCTFSelection is incorrect, and also the location. We should have more clear the division between xmipp and scipion.

- Its no clear the worth of this protocol (at least for me), since you can open any SetOfCTF, sort and select in an easy and friendly way. We need to keep in mind that maintaining this type of protocols costs time and human resources. I think there are alternative ways in scipion to do this that are pretty easy.

- Check that line length are less than 80, as proposed by the Python PEP8 style guide, that is adopted for Scipion.

josuegbl:
Im going to remove this PR since this changes  are in a PR to release-1.1-facilities.

PR #1003, opened 2017-05-14
Co kumar

cossorzano:


delarosatrevin:
@pconesa , I have the same suggestion here...these changes are in Xmipp code, if @rmarabini could not review them, @cossorzano could doble-check and merge the PR.

PR #1008, opened 2017-05-17
modify protocol_system_monitor so it plots gpu usage

rmarabini:


delarosatrevin:
I would recommend to "squash" this all commits with the same naming...this just mess up the commits history. 

Following a nice and clear blog about how to do it:
https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/

rmarabini:
Should this be closed since it has been replace by monitor_cuda?

PR #1009, opened 2017-05-18
Tm gain

majtner:
Added movie gain in streaming
Added movie gain monitor
Small changes in other monitors


pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -114,31 +114,32 @@ def _loadInputList(self):
...
-                        for m in self.listOfMovies)
-        outputStep = self._getFirstJoinStep()
-
-        if newMovies:
-            fDeps = self._insertNewMoviesSteps(self.insertedDict,
-                                               self.listOfMovies)
-            if outputStep is not None:
-                outputStep.addPrerequisites(*fDeps)
-            self.updateSteps()
+        if isinstance(self.inputMovies.get(), SetOfMovies):
```
delarosatrevin: Why this condition? If one wants to skip the _checkNewInput in a given protocol, it can be implemented as empty

pyworkflow/em/protocol/monitors/protocol_monitor_summary.py
```diff
@@ -27,24 +27,37 @@
...
+    - CTF estimation
+    - Movie gain estimation.
     """
     _label = 'monitor summary'
     _version = VERSION_1_1
 
     def _defineParams(self, form):
         ProtMonitor._defineParams(self, form)
 
+        form.addSection('MovieGain Monitor')
```
delarosatrevin: Make sure that the summary monitor is still working even if not Gain-estimation protocol is used

pyworkflow/em/protocol/monitors/protocol_monitor_movie_gain.py
```diff
@@ -0,0 +1,318 @@
...
+        if not os.path.exists(fnSummary) or os.path.getsize(fnSummary) < 1:
+            return False
+        fhSummary = open(fnSummary, "r")
+        if not os.path.exists(fnWarning):
+            fhWarning = open(fnWarning, "w")
+        else:
+            fhWarning = open(fnWarning, "a")
+
+        line = fhSummary.readlines()[-1]
+        values = line.split()
```
delarosatrevin: Using float(values[index]) everywhere is not very readable, apart from converting to float all the time. 
An alternative could be:

x, y, z, y = map(float, line.split()) # or whatever names fit for the case

and then use these float variables in the code.

pyworkflow/em/protocol/monitors/protocol_monitor_movie_gain.py
```diff
@@ -0,0 +1,318 @@
...
+    def getData(self):
+        prot = self.protocol
+        fnSummary = prot._getPath("summaryForMonitor.txt")
+        fhSummary = open(fnSummary, "r")
+        idValues = []
+        stddevValues = []
+        ratio1Values = []
+        ratio2Values = []
+        for idx, line in enumerate(fhSummary.readlines()):
+            values = line.split()
```
delarosatrevin: same, here...see previous comment about using: map(float, line.split())

pyworkflow/em/protocol/monitors/protocol_monitor_movie_gain.py
```diff
@@ -0,0 +1,318 @@
...
+    def _defineParams(self, form):
+        # ProtMonitor._defineParams(self, form)
+        form.addSection(label='Input')
+
+        form.addParam('inputProtocol', params.PointerParam,
+                      label="Input protocol", important=True,
+                      pointerClass='XmippProtMovieGain',
+                      help="This protocol will be monitorized")
+
+        form.addParam('stddevValue', params.FloatParam, default=0.04,
```
delarosatrevin: Check here that line length are less than 80, as proposed by the Python PEP8 style guide, that is adopted for Scipion.

config/templates/execution.summary.template.html
```diff
@@ -202,6 +202,10 @@
                 <H2>CTF monitor</H2>
                 <DIV id="ctfChart"></DIV>
             </SECTION>
+            <SECTION id="movieGain">
```
delarosatrevin: Make sure that the HTML report still looks fine even if the Estimate-Gain is not used.

pyworkflow/em/protocol/protocol_movies.py
majtner: _checkNewInput makes sence only if more than one movie is processed and, for example, in XmippProtMovieGain, input could be either SetOfMovies or Movie, so it can not be implemented as empty

pyworkflow/em/protocol/monitors/protocol_monitor_movie_gain.py
majtner: Here is problem that the line looks like this "movie_000001: 0.006494 0.988146 1.011829 1.440612", so there is a string before floats.
I understand the problem, I will try to correct it.

delarosatrevin: You can just convert from the second (index1) value by using list indexing:
x, y, z, y = map(float, line.split()[1:])

pyworkflow/em/protocol/protocol_movies.py
delarosatrevin: This condition seems not to be very problematic anyway, but in a more general context, it is better not to modify a base class to accommodate the particularities of subclasses. In this case, the inputMovies parameter is supposed to be a SetOfMovies. In that way, this protocol is "breaking" the inheritance rule and allowing to input also a single Movie object. Then, the way should not be changing the base class (what could break other protocols that depends on it) just for this case (it does not means that in some cases a re-design of the base class is required). 
Still the _checkNewInput in this protocol could be implemented as:
def _checkNewInput(self):
    if isinstance(self.inputMovies.get(), SetOfMovies):
           ProtProcessMovies._checkNewInput(self)

delarosatrevin: Identation has been messed-up with this wiki format.

pconesa:
I'm cancelling this, I'm moving this changes to facilities. Will PR soon.

PR #1028, opened 2017-05-26
Release 1.1 headless

pconesa:
This may not end up in release-1.1, but will serve us to compare and refine the code.
I'll annotate here what we discuss with @kevinpetersavage. Also, @delarosatrevin supervision is more than welcome.


matplotlibrc
```diff
@@ -0,0 +1 @@
+backend      : PDF
```
pconesa: This can be placed only in Diamond installation or use other ways: 
http://matplotlib.org/users/customizing.html#the-matplotlibrc-file

pyworkflow/em/packages/ispyb/protocol_monitor_ispyb.py
```diff
@@ -70,7 +70,7 @@ def monitorStep(self):
                                monitorTime=100)
 
         monitor.addNotifier(PrintNotifier())
-        monitor.loop()
+        monitor.step()
```
pconesa: All this should be ok.

pyworkflow/em/protocol/protocol_import/micrographs.py
```diff
@@ -382,12 +382,76 @@ def decompress(program, args, ext, nExt):
...
 
         dim = dimMovie.getDim()
+
         range = [1, dim[2], 1]
 
         movie.setFramesRange(range)
         imgSet.setDim(dim)
         imgSet.setFramesRange(range)
 
+    def iterFiles(self):
```
pconesa: This is the part that allowed importing frames when not streaming. This has been fixed by @delarosatrevin and is in release 1.1. So my guess is that this can be safely removed and it'll still work. Something to test.

delarosatrevin:
Ok, @pconesa I can take a look and we could discuss it on Monday or Tuesday.

pyworkflow/protocol/launch.py
```diff
@@ -213,18 +213,25 @@ def _submit(hostConfig, submitDict):
         print "** Couldn't parse %s ouput: %s" % (gcmd, redStr(out)) 
         return UNKNOWN_JOBID
 
-    
+def _pass_though_no_gui_state(command):
+    if 'SCIPION_NOGUI' in os.environ:
+        return 'export SCIPION_NOGUI=true;' + command
```
pconesa: I can't see where this env variable is read..am I missing anything here or is not needed?

matplotlibrc
pconesa: Or it can stay if we never merge this branch.

PR #1034, opened 2017-05-31
Monitor GPUs

pconesa:
This is meant to replace the other PR: https://github.com/I2PC/scipion/pull/1008
@rmarabini , could you check if we are not losing any change. Surprisingly, if you compare both branches the are not the same. But if you compare against devel seems to be both the same. I'm confused.

I was unable to do a squash as suggested by @delarosatrevin.


rmarabini:
Pablo, should I check rm_cuda or devel?

   Roberto

rmarabini:
I added a couple of extra functionalities, now it check network and disk usage.

PR #1079, opened 2017-06-26
Fix creation of volume from B-factor viewer

pconesa:


pyworkflow/em/packages/xmipp3/viewer_resolution3d.py
```diff
@@ -203,9 +203,9 @@ def _applyBfactor(self, e=None):
         newVol.setLocation(volOut)
         volSet.append(newVol)
         volSet.write()
-        
+
+        setattr(volSet, "protocol", self.protocol)
```
delarosatrevin: @pconesa Why do you need to set the protocol to the volSet? Be careful to no create cyclic dependencies. Usually the protocols are the top level containers and the outputs are "childs"

pyworkflow/viewer.py
```diff
@@ -202,24 +202,31 @@ def objectView(self, filenameOrObject, **kwargs):
...
-            **kwargs: Can receive extra keyword-arguments that will be passeed
+            **kwargs: Can receive extra keyword-arguments that will be passed
                 to the ObjectView constructor
         """
         # We can not import em globally
         from pyworkflow.em import ObjectView
         fn = None
 
+        def getProtocolId(obj):
+            if not hasattr(obj, 'protocol'):
```
delarosatrevin: I don't see the need to set the obj.protocol in one place and retrieve it here...since when the object does not have objId, always the viewer.protocol should be used and it does not change

pyworkflow/viewer.py
```diff
@@ -202,24 +202,31 @@ def objectView(self, filenameOrObject, **kwargs):
...
+                raise Exception("self.protocol is not defined for this Viewer.")
+            return obj.protocol.strId()
+
         if isinstance(filenameOrObject, basestring):
             # If the input is a string filename, we should take the object id
             # from the protocol. This assumes that self.protocol have been
             # previously set
             fn = filenameOrObject
-            if not hasattr(self, 'protocol'):
```
delarosatrevin: I think this was just a protection...but for all ProtocolViewers, the self.protocol should be set. For me it seems more complicated the logic of the new version.

pyworkflow/em/packages/xmipp3/viewer_resolution3d.py
pconesa: Tecnically, volSet is not an output of the protocol. It is an isolated SetOfVolumes without a reference to the protocol. and the protocol does not know anything about the set.

pyworkflow/viewer.py
pconesa: I'll explore that...as you can see I'm not familiar with the viewers.

pyworkflow/viewer.py
pconesa: actually, I think if object is basestring...it is impossible to have a attribute of any name. Am I right?:

`myString = "Hola"
setattr(myString,"newattribute","extra")

  File "<input>", line 1, in <module>
AttributeError: 'str' object has no attribute 'newattribute'
`

Am I missing anything here?

pyworkflow/em/packages/xmipp3/viewer_resolution3d.py
delarosatrevin: True, but it is still confusing given our current data model, and not really needed in this case.

pyworkflow/viewer.py
delarosatrevin: Here the attribute is been requested from 'self', that is a viewer class, not string.

pconesa: Ok, I'm realizing now my mistake. the attribute if for the viewer....Bad refactoring!! grrrr.

delarosatrevin: No problem...for me the raising alarm was that I didn't see the need to change the base class to fix this particular issue in the B-factor viewer.

pconesa:
@delarosatrevin , good that you reviewed this! Thanks. Is it better now. I couldn't see the "return [viewer]" option you've mentioned before.

pyworkflow/viewer.py
pconesa: So, then the problem is on the b-factor side....in the past we where passing a filename, therefore it was using self.protocol to get the Id, but for some reason, this was changed and we are passing to the factorymethod (objectView()) a set of volumes instead....should we change that on the B-factor side and pass a filename instead?. I guess we still need to create the set, but pass the filename.

delarosatrevin:
You're welcome @pconesa. The viewer should not be returned, instead the viewer should return one or many "views" when a given object should be visualized. You can find a similar example (about displaying a temporary SetOfVolumes) in the relion viewer.py module.

delarosatrevin:
The factory method should be fine, but taking into account to pass the protocol (or protocol id, can't remember exactly) as a parameter, that will be used later by the showj to associate possible new objects registered.

PR #1080, opened 2017-06-27
Use the run.db timestamp to set the lastUpdateTime in the protocol.

pconesa:


pconesa:
Hi @delarosatrevin , an user (dani) has reported something similar to what you commented a while a go. He is running thing on a cluster. I think this should fix the issue....but it would be nice to test it in a cluster. Is it easy for you to test this there?

delarosatrevin:
Hum...could this be the bug that I mentioned that we observed in our cluster? The symptom was that some jobs launched from the GUI stay as "launched"....I just removed the validation of the timestamp for a quick-fix.

pconesa:
yep, sounds like it.....this fix is using the same source of time (run.db filesystem timestamp).

delarosatrevin:
It was not happen always, but I could try...another user in Denmark also reported in their cluster, so I think we may have several places to check if this fix solves the mentioned issue.

Issue #1081, opened 2017-06-27
Movie alignment protocols not working with compressed movies


delarosatrevin:
With the refactoring for streaming, now the alignment protocols are broken when using compressed (e.g. bz2) movies as input. 

The problem is that the name is change after the movie is decompressed and the expected micrograph name differs.

PR #1084, opened 2017-06-28
Release 1.1.1 bug compressed movies

delarosatrevin:
Processing of input movies in compressed format was broken (presumably with streaming changes)
The problem was that the decompressed movie name was used for some outputs and in other context the original name was used to create the output object. 

Now with this fix, the root (without the last two extensions) will be used for the output and for the set creation. 

This should fix issue #1081. So after this PR is tested and merged, the issue can be closed.

Moreover, a minor code formatting was done in data.py and removed some debug print.


pconesa:
closes #1081

PR #1090, opened 2017-06-30
tm_gain branch files moved to facilities

pconesa:


pconesa:
Hi @majtner, could you please help me to verify that this has all the gain changes (streaming mode).

pyworkflow/tests/em/protocols/test_protocols_xmipp_movies.py
```diff
@@ -461,10 +461,7 @@ def getArgs(self, filesPath, pattern=''):
     def test_estimate(self):
```
pconesa: Test fails! there is no gain image. I'm not sure if I'm missing anything here @majtner

pconesa:
Gain figure does not show until another figure (like ctf figure is shown)

config/templates/execution.summary.template.html
```diff
@@ -202,6 +202,10 @@
                 <H2>CTF monitor</H2>
                 <DIV id="ctfChart"></DIV>
             </SECTION>
+            <SECTION id="movieGain">
```
delarosatrevin: @pconesa Please make sure this will work/display nice even when the gain is not used.

majtner: I believe we tested all the combinations of all three monitors that are currently implemented. You choose, which monitor(s) you want to display and only chosen ones are in HTML report.

delarosatrevin: Great!

pyworkflow/tests/em/protocols/test_protocols_xmipp_movies.py
majtner: @pconesa Probably I forget to modify the test protocol here. There was a change in protocol, where you can assign gain image instead of estimating it.

majtner:
> Gain figure does not show until another figure (like ctf figure is shown)
@pconesa How did you simulate it? Where it does not show?
In Movie Gain Monitor, there is only one monitor and in my tests, it always appears, when analyzing results.
In Monitor Summary, there is always System monitor and if you attach this protocol to movie gain protocol, you can generate either individual monitors (system or gain) or generate HTML, where both are presented.
Therefore, I am not sure, where did you find this problem.

pyworkflow/tests/em/protocols/test_protocols_xmipp_movies.py
majtner: @pconesa Ok, I found this problem and there are 2 possibilities to fix it. There is a new parameter, which is called "useExistingGainImage" and by default it is set to True, which means that we do not estimate gain image but we use the existing one, which should be imported together with movies. There is an assumption that you import movies together with corresponding gain image, which is used and therefore we do not have to estimate it.
I am not an expert so I do not know, how often is this the case. If I set this parameter by default to False, all tests will pass.
So we can either modify the default value (it makes sense in case it is more likely that there is no gain image) or we can change this parameter value in the test.

josuegbl: Well, you can set the parameter to None in the form as default value, and forces the user to choose one, and change to False for the test. 
Anyway, am not sure the meaning and the use of this parameter, since if its True, you do not estimate the "gain" image, the same result if you dont execute the protocol... Im sure that Im missing something.... can you clarify this, please?

majtner: Motivation for this modification was from Jose Maria. He wrote:

"Essentially, besides Scipion importing frames already corrected for gain, it might be interesting to also import videos BEFORE gain correction. Of course, the gain image should also be imported.
If they have been gain corrected, "business as usual". But if they have not been, you have to decide if you correct them and then check on the residual gain as you do now. Of, instead, you do not correct for the gain, but then you will be estimating the real gain and not the residual gain."

If I understand it correctly, there are two steps: estimation and application. So you can either estimate the gain and apply it to input, or you can use gain image from external source and have also gain not-corrected images in Scipion.

josuegbl: Ok, I see your point, but I disagree a little with you. I agree that there are two steps (estimation and application), but this steps should be apply in a different way. First of all, the goal of the protocol is an estimation of a residual "gain" of the camera from a movie, just to verify that the camera is working properly over the whole session. You have two situations here: movies that are corrected with the real gain image, and movies that are not corrected. For the last ones, when you import it, you can set its gain image, and this information is stored in Scipion. At this point, for the first ones, the estimation is as usual. But in the other case, you can either apply first the gain to the movies and after that estimate your "residual gain" or you can estimate the gain without apply the "real" gain image. In resume: I see the parameter as a flag to allow the application of the real gain image (that is stored as a property of the SetOfMovies) BEFORE the estimation of the residual gain, avoiding the application of this new gain image to the set in all cases.
Of course, if you want apply the real gain image to the set, you need to verify if exists when the flag "useExistingGainImage" is True.

majtner: Ok, I think it would be good to ask Carlos Oscar for his opinion on this, since it was his idea to add it this way to protocol.

josuegbl:
@majtner, this PR has all modifications that has been discussed with @cossorzano last week?

majtner:
@josuegbl, I have modified the code based on discussion with @cossorzano but I still need to discuss issues about gain figure with @pconesa

josuegbl:
Ok, thanks ... let me know when you're done, to merge your changes to release-1.1-facilities-devel.

majtner:
@pconesa I modified movie gain monitor based on our last discussion and everything seems to be ok, so it is your turn to test it

PR #1107, opened 2017-07-07
fix extract movie particles

josuegbl:
This PR sould fix an error that occurs when the partcles from movies wants to be extract


pyworkflow/em/packages/xmipp3/protocol_extract_particles_movies.py
```diff
@@ -225,8 +222,7 @@ def _processMovie(self, movie):
                 #TODO: there is no need to write the frame and then operate
                 #the input of the first operation should be the movie
                 movieName = imgh.fixXmippVolumeFileName(movie)
-                print "Passing Ih: ", frame, movieName, frameName
-                imgh.convert(tuple([frame-1, movieName]), frameName)
+                imgh.convert(tuple([frame, movieName]), frameName)
```
delarosatrevin: tuple([frame, movieName]) is a weird way to have just: (frame, movieName)

delarosatrevin:
Test passed before and after my simple commit.

PR #1114, opened 2017-07-12
create and schedule project scripts

pconesa:
create_project.py script to create a project from a scipion workflow file (json)
schedule_project.py to schedule all protocols given a project name (already existing project)


delarosatrevin:
Seems fine to me.

PR #1120, opened 2017-07-14
Release 1.1.facilities readonlyproject

### LINKED ISSUE #1112: Scipion read-only
Investigate what changes are needed to support Scipion read-only.

yaizar:
closes #1112

### DIFF
--- a/pyworkflow/project.py
+++ b/pyworkflow/project.py
@@ -513,8 +513,6 @@ class Project(object):
                 time.sleep(0.5)
                 self._updateProtocol(protocol, tries + 1)
 
-
-
     def stopProtocol(self, protocol):
         """ Stop a running protocol """
         try:
--- a/pyworkflow/project.py
+++ b/pyworkflow/project.py
@@ -35,8 +35,6 @@ import time
 from collections import OrderedDict
 import datetime as dt
 
-from sqlite3 import OperationalError
-
 import pyworkflow.em as em
 import pyworkflow.config as pwconfig
 import pyworkflow.hosts as pwhosts
@@ -508,7 +506,8 @@ class Project(object):
                 protocol.setFailed(str(ex))
                 self.mapper.store(protocol)
             else:
-                if isinstance(ex, OperationalError):
+                if str(ex) == 'attempt to write a readonly database':
+                    print("Trying to open project as read-only...")
                     self.setReadOnly(True)
                 time.sleep(0.5)
                 self._updateProtocol(protocol, tries + 1)
--- a/pyworkflow/project.py
+++ b/pyworkflow/project.py
@@ -497,28 +497,23 @@ class Project(object):
             prot2.getProject().closeMapper()
             prot2.closeMappers()
 
-        except OperationalError as ex:
-            print("ERROR: %s, tries=%d" % (ex, tries))
-            print("Opening project as read only...")
-            self.setReadOnly(True)
-            if tries < 3:
-                time.sleep(0.5)
-                self._updateProtocol(protocol, tries + 1)
-
         except Exception as ex:
             print("Error trying to update protocol: %s(jobId=%s)\n "
                   "ERROR: %s, tries=%d"
                   % (protocol.getObjName(), jobId, ex, tries))
-            if tries < 3:
+            if tries == 3:  # 3 tries have been failed
+                traceback.print_exc()
+                # If any problem happens, the protocol will be marked
+                # with a FAILED status
+                protocol.setFailed(str(ex))
+                self.mapper.store(protocol)
+            else:
+                if isinstance(ex, OperationalError):
+                    self.setReadOnly(True)
                 time.sleep(0.5)
                 self._updateProtocol(protocol, tries + 1)
 
-        if tries == 3:  # 3 tries have been failed
-            traceback.print_exc()
-            # If any problem happens, the protocol will be marked
-            # with a FAILED status
-            protocol.setFailed(str(ex))
-            self.mapper.store(protocol)
+
 
     def stopProtocol(self, protocol):
         """ Stop a running protocol """
--- a/pyworkflow/project.py
+++ b/pyworkflow/project.py
@@ -70,6 +70,7 @@ class Project(object):
         self.shortName = os.path.basename(path)
         self.path = os.path.abspath(path)
         self._isLink = os.path.islink(path)
+        self.isReadOnlyFilesystem = False
         self.pathList = []  # Store all related paths
         self.dbPath = self.__addPath(PROJECT_DBNAME)
         self.logsPath = self.__addPath(PROJECT_LOGS)
@@ -185,7 +186,7 @@ class Project(object):
 
     def saveSettings(self):
         # Read only mode
-        if not self.isReadOnly():
+        if not self.openAsReadOnly():
             self.settings.write()
 
     def createMapper(self, sqliteFn):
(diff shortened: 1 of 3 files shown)


pyworkflow/project.py
```diff
@@ -35,7 +35,7 @@
 from collections import OrderedDict
 import datetime as dt
 
-import datetime
+from sqlite3 import OperationalError
```
delarosatrevin: Even if we are using sqlite now for the underlying database, I would like to keep the project.py more agnostic of that. I don't envision a change of the db backend in the near future, but who knows.

pyworkflow/project.py
```diff
@@ -507,6 +508,8 @@ def _updateProtocol(self, protocol, tries=0, checkPid=False,
                 protocol.setFailed(str(ex))
                 self.mapper.store(protocol)
             else:
+                if isinstance(ex, OperationalError):
```
delarosatrevin: Why to set readOnly if there is something failing? Do we really get the notification of the error? I guess in some cases should be other reasons and we do not want to "hide" the problem by setting to readOnly.  I think if the is an scenario where the project should be set as readOnly, this should be done using the API in that context...but not here at this level.

pconesa: Actually, we should be more precise, maybe checking the "error message"? OperationalError could be for other reasons. But, @delarosatrevin, we are trying to prevent a failure when loading a project that is in a read only file system. Or at least the project folder.

yaizar: We set readOnly because this error happens for this reason (we try to open a project for which location we don't have write permissions and we get `OperationalError: attempt to write a readonly database` and the project never opens). Yes we really get the notification of the error the same as we did before (the print message has not changed), but now that you say it we should probably notify that we're opening as read-only too. Can you be more specific regarding in which context should we check this? Would it be preferable to check if we're in a read only case when we load the project settings? Like [here](https://github.com/I2PC/scipion/blob/master/pyworkflow/project.py#L243) ?

delarosatrevin: My concern is only how the issue is handled. Because imagine that for some reason an user is trying to load a project from read-only folder. Then is this an error or not? This depends on the circumstances, maybe the folder is read-only by mistake and the user (or admin) should fix that...or maybe is really a read-only project to be used as example. In any case, I think that we should not silently open it in read-only, because the user could try later to run some operations an d if it is not clear the issue, could looks like Scipion is not working properly. I would propose that in the GUI, when loading a project...it can be checked in advance if the project folder is read-only...or more general, if the user has not writting permissions...and then show a message saying that the project will be load in read-only mode because of that reason. But this is one level up and not in the internal implementation of project. Would that way help in the problem that you are facing?

yaizar: I'm not sure, aren't the project settings  loaded in that line I mentioned before? how can I set the project settings in the GUI before they're loaded in the project context? Should I write the value of readOnly in the settings file directly? It would make more sense to me to handle this using the exact error message like @pconesa suggested and leaving the GUI alone.. maybe letting the user decide if it is desired to open the project in readonly mode with a message alert instead of printing it in the console if we wanna be more explicit..

delarosatrevin: I mentioned the GUI because it is the point where the user interact with a project...but my point applies if I'm a "user" from the API and I'm trying to load a project. The point here is if we want to considering loading a project from a read-only folder as an error o not. I don't like the behavior when we take decisions without making explicit what should be happen. 

Anyway, I think there is a problem with the solution proposed in this PR, the method Project.setReadOnly will try to write in the settings.sqlite file, but if it is read-only this will fail (I have not test this) and will be in a failing loop. Maybe we should discuss more about this during my next visit to Madrid in a week.

yaizar: Sure we can discuss when you're here. That error does occur but it doesn't prevent the user from opening and exploring the project. Whenever the user tries to do something that requires writing it gets the error window saying that we're attempting to write a readonly database (very explicit). Also this solution only changes the behavior when the project we're opening has some running protocols (if there are no running protocols the project opens, so decisions are already taking place without making explicit what happens at the very beginning).

pconesa:
I see, you mean to check the permissions, further before...at project 
loading from any GUI code, setting the read only flag then?

Still, if used as API .....this still will be an issue...what about at 
project loading time?

pyworkflow/project.py
```diff
@@ -1221,6 +1229,9 @@ def isReadOnly(self):
 
         return self.settings.getReadOnly()
 
+    def openAsReadOnly(self):
```
delarosatrevin: openAsReadOnly is a bit confusing name...openXXX indicates more an action than a property. I'm not sure if we could just simply change isReadOnly implementation to consider both if it is set in setting.sqlite or if the folder is read-only (as the implementation of this function)

pyworkflow/project.py
```diff
@@ -72,6 +70,7 @@ def __init__(self, path):
         self.shortName = os.path.basename(path)
         self.path = os.path.abspath(path)
         self._isLink = os.path.islink(path)
+        self.isReadOnlyFilesystem = False
```
delarosatrevin: I will propose to encapsulate this property through a function call if it is going to be used as public API, if not...to add the _ prefix (I know that this is not applied to other attributes/methods, but we never is too late to improve)

Moreover, maybe isInReadOnlyFolder is a bit more accurate name...but just a proposition.

pyworkflow/project.py
```diff
@@ -220,6 +219,14 @@ def load(self, dbPath=None, hostsConf=None, protocolsConf=None, chdir=True,
...
                 "Cannot load project, path doesn't exist: %s" % self.path)
 
+        # If folder is read only, flag it and warn about it.
+        if not os.access(self.path, os.W_OK):
+
+            self.isReadOnlyFilesystem = True
+            print("Warning: don't have write permissions for project folder. "
+                  "Opening as READ-ONLY.")
+
+
```
delarosatrevin: What about if the user does not have neither read (r) or execution permission (x) on this folder? There is not much to do, but since we are validating this, we could fail gracefully for these cases.

pconesa:
@delarosatrevin , @yaizar , I've moved the warning inside the project load (API), why not, is just a warning.
And I've added "READ ONLY" after the project name in the project window (GUI). I'd like to push this changes to diamond branch.

delarosatrevin:
@pconesa Take a look on my comments...I'm sorry to request (not much) more work...but I strongly believe that we need to be very careful with the API and naming conventions from now on.

pconesa:
No worrries, I understand naming are confusing. I'll rename things. Thanks for reviewing on sunday!.

pconesa:
To avoid to much renaming...what about "openedAsReadOnly" instead of openAsReadOnly? I think it is more expressive to what has happened.

delarosatrevin:
OK, "openedAsReadOnly" describes better as you said.

pyworkflow/project.py
pconesa: upss. this is true, we need +x on the folder...umm

pconesa: @kevinpetersavage , do EM scientist have execution rights on the processing folders?

pconesa: With no execution permission we can not even do: os.path.exists(absDbPath)...this is a challenge!

delarosatrevin: Don't spend too much time on this @pconesa. My comment was just about to fail "gracefully", not more. ;)

PR #1122, opened 2017-07-14
Monores no mask

Vilax:
Now the mask is created in the protocol

python changes --> mask creation
c++ changes -----> to avoid false mask voxels


pconesa:
@Vilax how is this releated to the other PR? #1144

Vilax:
These changes are independent from the viewer.

PR #1137, opened 2017-07-21
Fixed relion classes distribution plot for 3D classification case

### LINKED ISSUE #1129: relion  classification  "Plot class distribution over iterations"
I think this is for @delarosatrevin ;-). I have seen that relion  classification has now a nice overall section that includes  the option "Prot class distribution over iterations" (Prot should be Plot?). For 2D works great but for 3D I get the error:

---
  File "/usr/local/scipion/pyworkflow/em/packages/relion/viewer.py", line 392, in _plotClassDistribution
    i, fn = row.getValue('rlnReferenceImage').split('@')
ValueError: need more than 1 value to unpack
----

The problem seems to be that the 3D classes are not stored in a stack (as the 2D case) but in several files

I fixed this first problem with the following code:

index=0
for loop.......

                rlnReferenceImage = row.getValue('rlnReferenceImage')
                if rlnReferenceImage.find('@')!=-1:
                    i, fn = rlnReferenceImage.split('@')
                    #index = int(i)
                else:
                    fn = rlnReferenceImage
                index = index + 1

unfortunately this breaks the plotting code

AssertionError: incompatible sizes: argument 'height' must be length 10 or scalar
malformed bucket chain in Tcl_DeleteHashEntry.

I understand that height (in the code this is assigned to a variable called values) should be a list with 10 entries (the number of iterations) but it is not the case.
Can you check what is wrong?

delarosatrevin:
Fixes #1129

### DIFF
--- a/pyworkflow/em/packages/relion/viewer.py
+++ b/pyworkflow/em/packages/relion/viewer.py
@@ -34,7 +34,8 @@ from pyworkflow.em.data import SetOfParticles, SetOfImages
 from pyworkflow.em.plotter import EmPlotter
 from pyworkflow.protocol.constants import LEVEL_ADVANCED
 import pyworkflow.protocol.params as params
-from pyworkflow.viewer import Viewer, ProtocolViewer, DESKTOP_TKINTER, WEB_DJANGO
+from pyworkflow.viewer import (Viewer, ProtocolViewer, DESKTOP_TKINTER,
+                               WEB_DJANGO)
 
 from protocol_classify2d import ProtRelionClassify2D
 from protocol_classify3d import ProtRelionClassify3D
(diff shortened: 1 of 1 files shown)


rmarabini:
Hi @delarosatrevin ,

  I try this pull request. Now I cat display the classes in 3D ;-) but  last two classes has the same  color. 
(see attached image)

In detail: my data set has 3 classes. Class 1 is painted in blue classes 2 and 3 are painted in red. 
By the way, the values of the variable classId are 1,2 and 3. Is this OK or should be 0, 1 and 2? 

figure_1

delarosatrevin:
@rmarabini I have added +1 to the code that is generated the color range...seems to be working better now for few classes. Classes 1, 2, ...N should be fine.

image

rmarabini:
meged, should I delete the brach?

delarosatrevin:
Yep!

rmarabini:
branch deleted

PR #1141, opened 2017-07-21
Release 1.1.facilities dash in filename #1073

### LINKED ISSUE #1073: Scipion fails when the filemane has " - "
It is interpreted a a parameter in xmipp programs and may be others

Possible solutions:

escape special characters
pass file names as between ""
...

### LINKED ISSUE #1098: Streaming end token file
Shows up in the streaming tab of import movies but I don't think it's  implemented. 
Should we implement it or remove it? The expected behavior of this param would be to enter a filename, say movie21.mrcs, so that the protocol stops when this movie is imported?

yaizar:
closes #1073 
closes #1098

### DIFF
--- a/pyworkflow/em/packages/xmipp3/protocol_movie_correlation.py
+++ b/pyworkflow/em/packages/xmipp3/protocol_movie_correlation.py
@@ -115,8 +115,8 @@ class XmippProtMovieCorr(ProtAlignMovies):
         inputMd = os.path.join(movieFolder, 'input_movie.xmd')
         writeMovieMd(movie, inputMd, a0, aN, useAlignment=False)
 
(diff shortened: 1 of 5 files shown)


delarosatrevin:
I think the work around in this PR still does not fix the import problem. I have found a couple of problems:
1) If no matching, empty sets are created (probably this happened before the changes of this PR)
2) The is not matching of all items for names like: mic1.mrc, mic2.mrc and mic10.mrc

I have added a failing test case for it.
Still, I think there is no easy solution for this problem without requiring from the user a more string naming matching (now is more flexible and allow partially names matching). One possible solution could be to consider longer names first, but still some wrong match could happens if not all CTF files are present for import.

PR #1149, opened 2017-07-24
Protocol ctf selection updated

ajimoreno:
In protocol_ctf_selection several new input parameters have been included.
Moreover, the protocol now generates several micrograph subsets as outputs, the main one (as previously) with the discarded mics labeled as "disabled", and two mic subsets (these are the new ones) the first one with only the accepted mics and the second one with only the discarded mics.


pconesa:
@josegutab, could you please review this?

delarosatrevin:
Did you mean @josuegbl?

pconesa:
yes! I keep choosing josegutab....thanks. grrr

PR #1158, opened 2017-07-31
fix mathing files issue in ctf import

### LINKED ISSUE #588: import ctf issue
1. import micrographs with names img1.mrc, img2.mrc ... img3050.mrc
2. put in one folder results from gctf or ctffind3/4 i.e. for a single mic: img21.ctf, img21_EPA.log, img21_ctffind3.log
3. run import ctf protocol with pattern "*ctffind3.log" - 1 matching file

I get ctfs and mics with 2 (!) items each, for img2 and img21. CTF values are the same, but micrographs are correct. I have added few prints into ctf import code and the output gives:

```
00012:   Matching files:  1
00013:   CTF for micrograph id 1 was not found. Removed from set of micrographs.
...
01124:   fileId= None mic.getObjId()= 1112 fileName= /home/igbmc/gsharov/scratch/test/img21_ctffind3.log micName= img2.mrc micBase= img2
...
01236:   fileId= None mic.getObjId()= 1224 fileName= /home/igbmc/gsharov/scratch/test/img21_ctffind3.log micName= img21.mrc micBase= img21
...
```

The problem is that [here](https://github.com/azazellochg/scipion/blob/devel/pyworkflow/em/protocol/protocol_import/ctfs.py#L126) in `micBase in fileName or micName in fileName` the first condition returns true, when we compare **img2** with **/home/igbmc/gsharov/scratch/test/img21_ctffind3.log**

yaizar:
closes #588


delarosatrevin:
I think the work around in this PR still does not fix the import problem. I have found a couple of problems:
1) If no matching, empty sets are created (probably this happened before the changes of this PR)
2) The is not matching of all items for names like: mic1.mrc, mic2.mrc and mic10.mrc

I have added a failing test case for it.
Still, I think there is no easy solution for this problem without requiring from the user a more string naming matching (now is more flexible and allow partially names matching). One possible solution could be to consider longer names first, but still some wrong match could happens if not all CTF files are present for import.

PR #1196, opened 2017-09-11
Commented out relion-2.1 installation

ldelcano:


delarosatrevin:
Hum... @ldelcano, @pconesa How we can install Relion 2.1 now?

ldelcano:
I uncommented it out because it was not working (no tgz for relion 2.1).

pconesa:
@josuegbl has not deployied the tgz to our servers yet. So before it was impossible and our test server was failing. Once we deploy the 2.1 tgz we will restore this,

PR #1197, opened 2017-09-12
added zernike moments as new 2D features

majtner:
- added Zernike moments as new feature extractor to classify_extract_features.cpp
- added parameter [--noDenoising] for turning off implicit denoising


majtner:
I optimized this code a little bit, based on our discussion with @cossorzano

PR #1199, opened 2017-09-14
Release 1.1.1 to facilities

pconesa:
@I2PC/devel, @I2PC/external 
We are going to move everyting in release-1.1.1 to release-1.1.facilities-devel, which will be the next "hotfix/minor release" branch. Sorry for going out of a standard gitflow flow.

In summary:
hotfixes should go to release-1.1.facilities-devel
new developments  to devel (as usual)


pconesa:
CAn't do this way, both are protected.

PR #1207, opened 2017-09-18
Open project without settings.sqlite

### LINKED ISSUE #1075: Tolerate opening a project without settings.sqlite or with inconsistencies in it.
Users came wit a project without settings sqlite. they've copied one from another project but still there were some errors (selected protocol id was invalid) and project could not be loaded, after deleting some row project was able to load.

yaizar:
closes #1075


delarosatrevin:
Hi @yaizar, I have some questions regarding this PR...

* Why the feature of creating a default settings.sqlite is not merged to release-1.1.facilities-devel? I think it can be useful there as well (and this will go into devel at some point in the near future)
* How are the summary changes related to the settings?
* Why these new functions return a dictionary? This does not follow the convention of similar methods (warnings, validate) that return a list of string (maybe with some special tags)

pconesa:
@delarosatrevin , I got some answers....I decided not to put many things in the facilities branch and make changes in devel. Although it's true, people will benefit from this.

Summary changes is related in the sense that @yaizar tested also how tolerant was scipion if you provide "minimum" info like a single run folder (000123_xmipp_whatever..) and try to load it.Turns out that Scipion already toretate the absence of other Runs, providing you copy run.db into the project root folder as project.sqlite. We decided to visualize the problematic runs with a "danger" character. and therefore I akew @yaizar to implement a "generic" warning/inconsistencies feature, so each protocol can  annotate, what thing seems to be wrong (like a missing Run folder) and generically display it. It is true that we could have added another section and not reuse the "Summary"....but again, it was my decision to reuse that, since, rarely we will see these warnings.

I like more the dictionary. Using tag will tightly couple the info with the API. which I think we should not do it. So the key is the key to the warning (shown in bold in the GUI), and the value is a description...we could develop a model for this but decided not to complicate things.

As usual....your wise advice is very welcome, but here are our reasons! ;-).

delarosatrevin:
Hi @pconesa, thanks for the clarifications. Some comments below.

* I will propose to add the ability to generate the missing settings.sqlite
to the facility branch...since it add robutness and it is not a big change
anyway. I known an user in the US that will benefit from that...since for
some weird reason that file is deleted.
* I'm fine to reuse the summary to warn about these possible problems in
protocols.
* I understand the advantages of using a dict and separating the info from
the display. But it has the drawback that you are limiting the developer to
a predefined format. Apart from that, I would argue that consistency is
important and in this case would be better to keep the same convention as
the related methods.

It is not my intention to interfere with every single change and slow down
the development process, but I think is very important to be careful with
changes in the core and make them with consensus.

On Sep 18, 2017 5:50 PM, "Pablo Conesa" <notifications@github.com> wrote:

@delarosatrevin <https://github.com/delarosatrevin> , I got some
answers....I decided not to put many things in the facilities branch and
make changes in devel. Although it's true, people will benefit from this.

Summary changes is related in the sense that @yaizar
<https://github.com/yaizar> tested also how tolerant was scipion if you
provide "minimum" info like a single run folder (000123_xmipp_whatever..)
and try to load it.Turns out that Scipion already toretate the absence of
other Runs, providing you copy run.db into the project root folder as
project.sqlite. We decided to visualize the problematic runs with a
"danger" character. and therefore I akew @yaizar <https://github.com/yaizar>
to implement a "generic" warning/inconsistencies feature, so each protocol
can annotate, what thing seems to be wrong (like a missing Run folder) and
generically display it. It is true that we could have added another section
and not reuse the "Summary"....but again, it was my decision to reuse that,
since, rarely we will see these warnings.

I like more the dictionary. Using tag will tightly couple the info with the
API. which I think we should not do it. So the key is the key to the
warning (shown in bold in the GUI), and the value is a description...we
could develop a model for this but decided not to complicate things.

As usual....your wise advice is very welcome, but here are our reasons! ;-).

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/I2PC/scipion/pull/1207#issuecomment-330266482>, or mute
the thread

pconesa:
he,he. @delarosatrevin  no worries about your input. We need it and is important. And quite often you accept suggestions although not the way you  would have done them. THis is how collaboration works, right?.

I'll chat with @yaizar and try to accommodate your suggestions. Didn't know about the US user!. Good to know.

delarosatrevin:
Thanks for your understanding @pconesa! I don't want to create more
problems than solutions ;)

delarosatrevin:
Thanks @yaizar for addressing my suggestions.

pconesa:
I close this one since @yaizar made a PR to facilities-devel.

PR #1211, opened 2017-09-19
Open project without settings.sqlite

yaizar:


pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1211,6 +1211,8 @@ def _getNodeText(self, node):
                 nodeText = node.getName() + expandedStr
             else:
                 nodeText = nodeText + expandedStr + '\n' + node.run.getStatusMessage()
+                if node.run.summaryWarnings:
+                    nodeText += u' \u26a0'
```
delarosatrevin: This warn sign looks nice, but I'm always a little bit skeptic about weird systems and fonts...if still will be displayed nicely or will throw some error. It will also put at the interface level, not here...if in the GUI there is an style to visualize some kind of nodes (depending of some factors), it make more sense there than here in the info about the node.

delarosatrevin:
It works great! I just commented a little bit about a special character added to the node text...apart from that, the PR can be merged.

PR #1222, opened 2017-09-22
Fix picking threshold

### LINKED ISSUE #1164: xmipp picker issue
1.1.1 branch
Xmipp supervised picker seems not to take into account that I changed the threshold after training run. It 
 saves all autopicked particles (435) instead of 316+27
screenshot_20170804_190229

pconesa:
Well, still there is some work to be done, but I think is now fixed #1164 plus can be of any use.

Closes #1164


pconesa:
My bad, wrong base.

Issue #1221, opened 2017-09-22
Mark worse image after automatic picking


cossorzano:

pconesa:
Can you elaborate this more, please? @cossorzano

pconesa:
Make the worse automatic picked image when loading a new micrograph in the picking

PR #1223, opened 2017-09-22
Threshold in manual picking

### LINKED ISSUE #1164: xmipp picker issue

pconesa:
Well, still there is some work to be done, but I think is now fixed #1164 plus can be of any use.

Closes #1164

Some Threads  are now named
Change color of automatic particles from darker to a 0.33 hue.
Trigger save when threshold changes.
Show right count of automatic particles when moving the threshold
Save automatic particles below threshold as deleted.
Generate coordinates without automatic particles below threshold
Load the chosen color in the colorChooser dialog.
DEV HACK: triple click in "Threshold" label will show unavailable particles (deleted and below threshold)
Remove unused methods.
Still not perfect but better than before. At least threshold could be useful.


azazellochg:
Hi @pconesa , I have tried to make it count worng but couldn't, great job! :) A couple of things:

- Help --> Tips: there is no shortcut for "hide/show particle markers" - maybe missing ?
- Help --> Online help returns error: Desktop doesn't support the browse action (fatal), while it could open default browser with https://github.com/I2PC/scipion/wiki/Picker page
- Good idea to highlight "Threshold" with the red color of auto-picked particles, but it's only highlighted when clicking on a new/not picked micrograph. Maybe it should be highlighted all the time or if there are autopicked particles in a certain mic?

pconesa:
Many thanks @azazellochg. The red color means that threshold has changed. I see is not clear. Actually it was more for myself, but decided to leave it. Umm.

I think there are some improvements to make in terms of visualization: e.g, show deleted, show hidden due to threshold,...even remove the deleted ones...but had to stop at some point.


As for the shortcuts......he,he "Space" is the key. I'll write "Space". and fix the browser error.

pyworkflow/em/packages/xmipp3/convert.py
```diff
@@ -35,6 +35,7 @@
 from collections import OrderedDict
 from itertools import izip
 import numpy
+import time
```
delarosatrevin: Why this import? I don't see where it is used, maybe some debug?

delarosatrevin:
Great job @pconesa. I just took a quick look at the Java code, but I did see anything problematic at a glance. I will just test a little bit more (as a user) and merge the PR.

delarosatrevin:
@pconesa I tested and it seems fine. 

I just have two comments:
* I don't like much the red color for the coordinates (center, box, circle). Red is usually associated to bad things or errors. For the threshold is helps to highlight. Any other color proposal for the added particles?
* It is not related to this fix, but I would ask you if you can remove the Question after correcting the training (adding or removing particles) that ask if you want to use the corrections for the train....I think it does not make sense to do the training and not use the changes for the correction...for me it is quite annoying.

pyworkflow/em/packages/xmipp3/convert.py
pconesa: Removed! My bad.

pconesa:
@delarosatrevin, regarding the color....is based on a HUE change....I just move the HUE 0.33 from the picked color....I'll try to avoid the red by default.

Regarding the question......you want to remove the question, but do the calculations? Am I getting it?

pconesa:
What about this:
image

delarosatrevin:
Hi @pconesa, thanks a lot for considering this and I'm sorry for generating some extra work. 

Regarding the color, still too pinkie (redish) is it easy to have one closer to blue, for example?

Regarding the question, I mean to remove the question and do the update of the model (not sure if that is what you called calculations)...but is it what makes more sense (in my opinion) if you are in the training mode, I can't imagine a case where you are in training and want to proceed without correction the picking algorithm. I'm not sure if I have added some light to my comment or not. ;)

pconesa:
Yes you are clear.....actually while doing it I thought the same...but since the question was there, I thought someone though it was useful to ask.

As for the color.....that is easy....
image
I'll leave it blue, by default...but please, be aware that if you chose another color, the "automatic color" may fall on the red side.

delarosatrevin:
Ok, great! Thank you so much. Let me know when you submit the changes to update my local version, test it (just a little bit more) and merge the PR.

pconesa:
Clear like water! ;-)

pconesa:
Hi @delarosatrevin, I've done the changes you suggested plus #1221 and some tweaks.

pconesa:
Closes #1221

PR #1224, opened 2017-09-24
Showj fixes when visualizing .star files

delarosatrevin:
Changes in this PR:
* In showj (MetaDataGallery class) now the Relion rlnAnglePsi is multiplied by -1 to be understood as Xmipp convention.
* When exporting stack of images and applying the geometry...and using a star file...change the Relion labels to the expected Xmipp ones (and multiply by -1 as well rlnAnglePsi)
* For renaming the columns, the 'renameColumn'  function was added to the binding.

Still, there is a bit confusing behavior when showing classes 2D:
* If open the images of a class, using "Open images" it shows fine
* If open the images of a class, from the combobox, it will not apply the geometry by default (since the classes block does not have geometry)

### DIFF
--- a/software/em/xmipp/java/src/xmipp/jni/MetaData.java
+++ b/software/em/xmipp/java/src/xmipp/jni/MetaData.java
@@ -488,6 +488,9 @@ public class MetaData {
 	/** Add new column to MetaData */
 	public native void addLabel(int label);
 
+	/** Rename a column off the MetaData */
+	public native void renameColumn(int oldLabel, int newLabel);
+
 	/** Get the average and std images, result is left on input image */
 	public native void getStatsImages(ImageGeneric imageAvg,
 			ImageGeneric imageStd, boolean applyGeo, boolean wrap, int label);
@@ -512,6 +515,8 @@ public class MetaData {
 	 * @
 	 */
 	public native void fillConstant(int label, String value);
+
+
 
 	/**
 	 * Fill column in a random way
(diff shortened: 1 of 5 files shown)


software/em/xmipp/java/src/xmipp/jni/MetaData.java
```diff
@@ -488,6 +488,9 @@ public boolean getEnabled(long objId) {
 	/** Add new column to MetaData */
 	public native void addLabel(int label);
 
+	/** Rename a column off the MetaData */
```
pconesa: of not off?

delarosatrevin: Nop, is just to rename the Column named "off"

pconesa:
Hi, I've been looking at this with  @josuegbl and Jaime. The Psi (angle) issue is corrected...but there seems to be a problem with shifts, not reported before. Based on Jaime.....shiftsX and Y need to be swap and inverted (only X) to get the right results.......@josuegbl will look at the exact problem and get back.

pconesa:
More details....working with TestRelionClassify2D project generated by our tests.
a.- This is a subset (particle for class 3)  of the 2D relion classification protocol
image

b.- This is the particles.sqlite of the subset protocol with only those particles. Note that now the angle is properly understood (fixed).
image
Nevertheless, the average does not match, is more blurry.

c.- If I add a new relion 2d classify and show the star file that serves as input for relion for those same 18 particles:
image

Note the the values of the first star file and the last one should be exactly the same if we do not want to lose accuracy. Right?

delarosatrevin:
@pconesa @josuegbl Any progress on this?

PR #1225, opened 2017-09-24
Rm resolution fit

rmarabini:
Hi @delarosatrevin ,

This is a new branch related with the introduction of resolution and fitQuality as parameters of the ctfModel.

./scipion test --grep ctf  --run

runs properly including program_stream_monitor

./scipion test tests.em.protocols.test_protocols_xmipp_mics.TestXmippExtractParticles

fails in a different protocol (after extracting the particles). I do not think the fail is related with the change


delarosatrevin:
You are right @rmarabini , test are passing now, but still the extract particles is failing.
@josuegbl Can you take a look at this?

josuegbl:
ok, I put it into my queue :)

pconesa:
@rmarabini , previous changes (addition of resolution) is causing the 
`scipion test tests.em.data.test_convert_xmipp.TestSetConvert`
fail, I think test files do not have the new "_resolution" field.

rmarabini:
OK, I will check.

PR #1232, opened 2017-09-29
Model 01 import

mmmtnez:
Dear All,

   We are going to break the very large modelbuilding pull request in many smaller pull request.s
This first pull request deals with "volume import".

1) Now users may provide a origin when importing a volume
2) volumes may be visualized with chimera. In this case the origin information is taken into account

Test:
./scipion test tests.em.protocols.test_protocols_import_volumes.TestImportVolumes


delarosatrevin:
That's is a great news. I will take a look at this one.

pyworkflow/em/data.py
```diff
@@ -416,6 +416,14 @@ def __init__(self, location=None, **kwargs):
         # this matrix can be used for 2D/3D alignment or
         # to represent projection directions
         self._transform = None
+        #default orign by default is box center =
```
josuegbl: If _origin is used only for volumes, has more sense put this in Volume class? For the cases of 2d images, we have the Transform Matrix, that contain the shifts to center the particles.

pyworkflow/em/data.py
```diff
@@ -533,6 +541,28 @@ def getTransform(self):
...
+        return self._origin is not None
+
+    def getOrigin(self, returnInitIfNone=False):
+        if self.hasOrigin():
+            return self._origin
+        else:
+            if returnInitIfNone:
+                t= Transform()
+                x, y, z=self.getDim()
+                if z == 1:
```
josuegbl: I think this code is not optimized: something like: z = z/2 if z>1 it is better

pyworkflow/em/data.py
delarosatrevin: @mmmtnez (or should I said @rmarabini ?) We need also to be pragmatic. There are many assumptions, but still are have got some processing done so far. The arguments in favor of having _origin in the Image rather than Volume are not very convincing for me. Anyway, I'm fine with it as long as we have None by default. I'm always concerned about performance...more in not always better. Is there any plan to address the current performance issues that we already have identified in conversion routines?

josuegbl: @delarosatrevin, I spoke with @rmarabini about it.

pyworkflow/em/protocol/protocol_import/__init__.py
```diff
@@ -30,10 +30,11 @@
...
+from ctfs import ProtImportCTF
 from images import ProtImportImages
+from masks import ProtImportMask
 from micrographs import ProtImportMicrographs, ProtImportMovies
 from particles import ProtImportParticles, ProtImportAverages
-from coordinates import ProtImportCoordinates
-from ctfs import ProtImportCTF
 from volumes import ProtImportVolumes, ProtImportPdb
-from masks import ProtImportMask
+from viewer_import_volume import viewerProtImportVolumes
```
josuegbl: Maybe change the name of the file  viewer_import_volume.py to viewer.py is better.

pyworkflow/em/protocol/protocol_import/viewer_import_volume.py
```diff
@@ -0,0 +1,172 @@
...
+            count += 1
+        if len(_setOfVolumes) > 1:
+            f.write('tile\n')
+        else:
+            f.write("volume#1 originIndex %d,%d,%d\n" % (x, y, z))
+        f.close()
+        import pyworkflow.em as em
+        return [em.ChimeraView(tmpFileNameCMD)]
+
+    def _showVolumesXmipp(self):
```
josuegbl: I dont like the name " _showVolumesXmipp", since this function is not in Xmipp. Maybe something like  _showVolumesSlices is better

pyworkflow/em/protocol/protocol_import/viewer_import_volume.py
```diff
@@ -0,0 +1,172 @@
...
+        # Write an sqlite with all volumes selected for visualization.
+
+        sampling, setOfVolumes = self._createSetOfVolumes()
```
josuegbl: Why SetOfVolumes is created when one volume is imported if you are showing the object Volume (self.protocol.outputVolume), not SetOfVolumes?

pyworkflow/em/protocol/protocol_import/viewer_import_volume.py
```diff
@@ -0,0 +1,172 @@
...
+    # =========================================================================
+
+    def _showVolumes(self, paramName=None):
+        if self.displayVol == VOLUME_CHIMERA:
+            return self._showVolumesChimera()
+
+        elif self.displayVol == VOLUME_SLICES:
+            return self._showVolumesXmipp()
+
+    def _getVolumeName(self, vol):
```
josuegbl: maybe something more general like: volFn = vol.getFileName().split(":")[0] is better

pyworkflow/em/protocol/protocol_import/volumes.py
```diff
@@ -25,135 +25,177 @@
...
+        vol.setSamplingRate(samplingRate)
         copyOrLink = self.getCopyOrLink()
         imgh = ImageHandler()
 
         volSet = self._createSetOfVolumes()
-        volSet.setSamplingRate(self.samplingRate.get())
+        volSet.setSamplingRate(samplingRate)
 
         for fileName, fileId in self.iterFiles():
             dst = self._getExtraPath(basename(fileName))
```
josuegbl: As you are importing volume(s), you can fix the name, according Xmipp standard for the cases that ends either .mrc or .map, using the function ImageHandler.fixXmippVolumeFileName(volFn).

mmmtnez:
Dear @josuegbl : 
We have followed your recommendations or justified why we can not follow them. 
Please, continue with the review process.

rmarabini:
Hi @delarosatrevin <https://github.com/delarosatrevin> and @josuegbl,

What I mean is the following: Default value of origin is None but if you
use the function getOrigin the value returned by that function when
origin=None is the box center. ;-)

PR #1234, opened 2017-10-02
Co max movie shift

dmaluenda:


pconesa:
I close it and will push one from facilities.

PR #1240, opened 2017-10-03
Facilities to headless

pconesa:
Updating headless with latest changes from facilities. 
File browser have shortcuts to common locations.


delarosatrevin:
Hum... @pconesa is looks to me that there are many changed files? Isn't it? I wouldn't expect that this branch should be that different. In any case, I think we should updated first headless with facilities-devel and then the other way around.

pconesa:
I'll review what I've done but I'm not surprised...we have done many thing the release-1.1.facilities...and...isn't this what I'm doing....facilities-devel -> headless (I used headless devel, because is what @kevinpetersavage asked and make sene...I think the headless is being stale for quite a long time.

delarosatrevin:
Ok, I just was surprised by some many changes...I thought that headess was
not so out-of-date from facilities-devel

pconesa:
I think current branch in use in DLS is headless-devel. @kevinpetersavage ?

pconesa:
I'm closing this since we agreed to reverse the "flow".

PR #1241, opened 2017-10-03
Facilities summary monitor improvements

### LINKED ISSUE #1213: facilities-devel: generate html report in background
monitor summary -> html report

If you press the button "HTML report" and you have (as is my case) 1000 processed movies the protocol will process the 1000 movies BEFORE returning scipion control. This action takes a very looooooong time (enough to write 4 new issues, check the mail and do many other things). After clicking "html report" you should get Scipion control back and if creating the files is going to take a long while the user should be warned

yaizar:
closes #1213


delarosatrevin:
@yaizar Have you checked if the thumbnail is already available? The motioncor2 protocol can generate the thumbnail...in which case does not make sense to do it again for the html report.

yaizar:
oh! I checked but not with motioncor2. Will do.

delarosatrevin:
I think you can check if the SetOfMicrographs items have a property
"thumbnail" (or something like that, I'm not sure about the name..take a
look at motioncor2 how it is set)

delarosatrevin:
@yaizar Coincidentally, I'm installing and testing Scipion for streaming in Aarhus...I have noticed that even if the gain protocol is not used, that section is still shown in the html report. Can we remove that section if this protocol is not used?

pconesa:
@delarosatrevin , yes you are right. Is this thing that was done, but mysteriously has disappeared.
We'll fix that.

delarosatrevin:
Yes, I remembered that you and Thomas said that was not shown. Git
mysteries? ;)

delarosatrevin:
@pconesa, remember the saying, "if you drink, don't git"

pconesa:
not mysteries, probably human mistakes!

delarosatrevin:
Well, it is always good to have something to blame on.

delarosatrevin:
image

It seems that we are getting a negative count on the micrograph coverage???

pconesa:
upss! I'll have a look...can you send/attach the HTML?

pconesa:
Gain section should not show up if there is no gain protocol and negative values should not show up in the coverage. Not sure if this last change will fix the 2 "X-axis displacement".

PR #1242, opened 2017-10-04
Spider filter wizard: warn if spider is not installed

### LINKED ISSUE #997: Filter particle Wizard does not warn if package is not installed.
I was using the filter wizard for spider changing values and nothing was happening.
Then I realized that I didn't had Spider installed. Once I installed it, wizzard was doing something.

pconesa:
Closes #997 

**Protocol**: have a isInstalled() getter default ro True, to keep behaviour.
**Environment**: setInstallationPath getter, to check if em package is installed
**Spider filter wizard**: warn if spider is not installed
**Viewprotocols**: Check if the protocol is installed and display accordingly  (for Search and three-All view): grey protocol icon and grey font.

Fixed some failing test: xmipp_workflow and xmipp_convert


pconesa:
@delarosatrevin , hi, I would appreciate your comments on the way I've done this.

pconesa:
Sample image:
image

pconesa:
If you agree on the way...I can extend this to all the packages, now is only done for Spider...for an easy review. In that case do not merge it until I finish.

pyworkflow/utils/utils.py
```diff
@@ -597,6 +597,19 @@ def addLibrary(self, libraryPath, position=BEGIN):
         else:
             print "Some paths do not exist in: % s" % libraryPath
 
+    def setInstallationPath(self, path):
```
delarosatrevin: Can we try to avoid defining more variables and logic?

pyworkflow/utils/utils.py
```diff
@@ -597,6 +597,19 @@ def addLibrary(self, libraryPath, position=BEGIN):
...
+        """ Path to the package installation"""
+        # noinspection PyAttributeOutsideInit
+        self._installationPath = path
+
+    def isInstalled(self):
```
delarosatrevin: Maybe we can figure out when a package is installed from the getEnvironment function? In the environment could be many path that we could check.

pyworkflow/protocol/protocol.py
```diff
@@ -303,6 +303,10 @@ class Protocol(Step):
     # Version where protocol appeared first time
     _lastUpdateVersion = pw.VERSION_1
 
+    @staticmethod
+    def isInstalled():
```
delarosatrevin: I think this is a useful function to validate in some cases before execution, but I would say that it is more at the package level, not at the protocol.

pyworkflow/em/packages/spider/protocol/protocol_base.py
```diff
@@ -80,3 +80,8 @@ def runTemplate(self, inputScript, ext, paramsDict):
             if 'FATAL ERROR ENCOUNTERED IN BATCH MODE' in line:
                 raise Exception('Spider script error!')
         f.close()
+
+    @staticmethod
+    def isInstalled():
```
delarosatrevin: This could be defined more generally for all protocols depending if the 'package' of the protocol is installed or not.

pyworkflow/utils/utils.py
pconesa: I think I don't get you....you mean like setting up a isInstalled setter straight from the environment "creation"?

pyworkflow/protocol/protocol.py
pconesa: I actually looked at a "package" entity...but couldn't find it. I guess I'm missing something here. BUt I agree, this is more a package level attribute.

pyworkflow/em/packages/spider/protocol/protocol_base.py
pconesa: Same here....do we have a "package" object to query having a protocol..something like prot.getPackage().isInstalled() ?

pconesa:
I've tried to reduce the logic and "move" the isInstalled to a "Package" context.
I've realized that all the protocols have a _package that is the "module". Didn't know that.

delarosatrevin:
@pconesa , Now it is better...but still I don't like to introduce another concept ("Package") to the developer of a new module. The package associated to a protocol is the Python module itself, not a new class defined by us. I would like if we can came up with a more general solution without adding extra overhead to future developers.

delarosatrevin:
Do we need to introduce this changes now for the release? The more I think about a solution, the more I think it should go to the "pluginization" where more "duties" will be moved to the packages (plugins). For example, the packages should know about the variables that should be added to the configuration...and these could be validated to exists as a general way to check if it is installed or not. But not there is not such a link in a clear way.

pconesa:
The existence of package is optional....if it exists it will be used to check if the package is installed. Do not take the class package too seriously.....but I relay think that sooner or later, we will need an entity (name it package, plugin,...) to concentrate data and logic for the pluginization. I was playing with the concept of package., but obviously I can remove it and have an "optional boolean module variable" instead? (is this what you have in mind?)

delarosatrevin:
I think the idea of having the "Package" class is very good, specially as you said for unifying data/logic for the pluginization. I just wanted to warn about making it automatic and as much simple as possible for other developers. But I think that would be better to introduce it after we have some discussions/brain storming about the pluginization. Maybe for now is it enough with a simple validation in the wizard. As it is done in other places?

pconesa:
It looks ok to me @delarosatrevin . Shall we do other packages? to be displayed in gray?

pconesa:
@delarosatrevin , this is done. Ready to review.

PR #1244, opened 2017-10-06
Fix parse gctf

delarosatrevin:
@azazellochg I has fixed a bug in parsing Gctf log files.
When the 'High resolution refinement' and not the 'Phase shift estimation' the line produced was as below:

23846.26    23525.42       61.84    0.521613  Final Values (Hres refine)

Adding the extra '(Hres refine)' string and making the number of elements after split greater...and broken the current parsing. 

I have changed the parsing a little bit and this error is fixed. Can you double check my changes to make sure it is working as expected? (I have passed the Gctf tests already)

### DIFF
--- a/pyworkflow/em/packages/gctf/convert.py
+++ b/pyworkflow/em/packages/gctf/convert.py
@@ -75,7 +75,7 @@ def parseGctfOutput(filename):
                 # Take ctfResolution as a tuple
                 # that is the last value in the line
                 # but remove escape characters first
-                result[5] = float(line.strip().split()[-1])
+                result[5] = float(line.strip().split()[-2])
                 break
         f.close()
     else:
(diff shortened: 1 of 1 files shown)


azazellochg:
@delarosatrevin I'll check it as soon as I can ;)

delarosatrevin:
Ok, great! Thanks for your help.

azazellochg:
@delarosatrevin it looks ok. The only thing I noticed is that defocus values are swaped and defAngle=defAngle(GCTF)+90. Was this always like this and I didn't pay attention?

delarosatrevin:
@azazellochg I only changed the way the values were parsed. The CTF defocus values are "standardized" according to the EMX definition. You can see the function here:
https://github.com/I2PC/scipion/blob/master/pyworkflow/em/data.py#L191

If it is working fine, we can merge and delete the branch.

azazellochg:
ah yes, forgot about this standardize function. Good to go!

PR #1245, opened 2017-10-06
Fix volume name before showing the filter volume wizzard

### LINKED ISSUE #1115: volume wizards issue
I might be missing something, but in 1.1 version the filter wizard shows incorrect image on the left panel - it looks like masked something... This is the case with all wizards working on volumes. When I look at the volume itself it looks fine.
screenshot_20170712_155500

pconesa:
Closes #1115


pconesa:
Hi, @azazellochg would you have time to verify this. This should fix #1115.

pconesa:
many thanks @azazellochg

Issue #1208, opened 2017-09-19
phase shift not written by 'relion - export ctf'


JuhaHuiskonen:
Protocol 'relion - export ctf' does not write the value of phase shift in the output STAR file.

azazellochg:
it is not exported for new projects, together with _resolution and _fitQuality

Issue #1206, opened 2017-09-18
phase shift not shown


JuhaHuiskonen:
Analyze results -table does not show the phase shift determined by CTFFIND4

PR #1248, opened 2017-10-10
Release 1.1.facilities relion fixes

### LINKED ISSUE #1243: relion protocols always use oversampling 1
Relion refine and classification protocols have '--oversampling 1' hard coded. Even if '--oversampling 0' is given from the command line, it is not effective.  It might be useful to construct the command line so that additional parameters come first - or add oversampling in the GUI as an advanced option the user could choose.

delarosatrevin:
Main changes:
* Added oversampling in extraParams, now user can change its value. (closes #1243 )
* Added the prior labels in input Relion .star files (when considering the previous alignment) 
* Added ctffind4_ctfPhaseShift when export relion CTF (closes issue #1208 )
* Display ctffind4_ctfPhaseShift by default if present (closes issue #1206)
* Restored to 50% when displaying CTF or Micrographs


josuegbl:
Is ok for me...

PR #1258, opened 2017-10-16
Added variance features

majtner:
Added variance features to xmipp_classify_extract_features
- feature vector consists of 17 values: 16 variance feature values (for 4x4 regions of a single particle) + 1 feature value as their combination that is used for eliminating false positives or the bad particles


majtner:
Based on discussion with @cossorzano, I moved, so-called, variance features from different local branch here

PR #1311, opened 2017-11-05
Protocol for eliminating false positive particle in streaming

majtner:
- added new protocol for eliminating false positives
  - using already existing variance features from classify_extract_features
  - protocol works also in streaming

- modified protocol_create_stream_data for handling also SetOfParticles
  - this enables streaming also particles, which is used in testing new protocol

- added new test TestXmippEliminatinfFalsePositives, which is testing XmippProtEliminateFalseParticles in streaming and also without streaming


cossorzano:
It would be better if the program is called xmipp_image_eliminate_empty_particles. It would also be great if the protocol produces two output sets of particles, one with the the selected particles and another one with the unselected particles.

majtner:
The original name of the protocol was eliminating empty particles. Based on our discussion with @pconesa and others during my presentation about this topic it seemed to me that "empty particles" are not good term, since it was confusing for some people. Therefore, I changed it to "false positives".
It is, of course, no problem to change it back but I would rather ask @pconesa if he also agrees, since he was suggesting this change.

And yes, the protocol has two output sets. The first one is by default "output.xmd" and the second one is "eliminated.xmd", where len(output) + len(eliminated) = len(input)

pconesa:
Sorry for introducing some entropy here....my point was that empty particles could mislead the users, but the final decision is yours, mine was just my "humble opinion", ;-).

majtner:
@pconesa I took your opinion from the user point of view, which is very relevant and therefore I opened it for discussion here, since I also agree that "empty particles" might be ambiguous. From my perspective, "false positives" is more precise but if everybody agrees with "empty particles", I will change the names back.

PR #1337, opened 2017-11-14
Shorting protocols labeling sequence

dmaluenda:


pconesa:
@dmaluenda works fine, well done!

PR #1347, opened 2017-11-20
Model 03 powerfit

mmmtnez:
Dear all,
as we said in our two previous pull requests, we divided the large model-building pull request in smaller pull requests. This third pull request (model_03_powerfit), coming after the 'extract unit cell' one (model_02_import), deals with the rigid fitting of an atomic structure and a density volume.

1.- The origin assigned to the  input volume is passed to powerfit program.

2.- A data set containing an atomic structure (pdb) and a map volume (mrc file) -useful for testing the protocol- has been added to the data folder (model_building_tutorial folder that contains two folders, PDBs and volumes).

3.- The coordinate system and the resulting rigid fitting between the input atomic structure and the density volume may be visualized with chimera.

4) PDB objects may contain a volume object, in this case if no volume is given, powerfit will use as input the assigned volume

5) powerfit may be installed using "scipion install" approach

6) the method adaptBinFileToCCP4 has been further refined. Given an input volume it will created -if needed- a mrc file with the proper header

Test:
./scipion test tests.em.protocols.test_model_building_workflow

Requirements:
install powerfit program


mmmtnez:
Please, ignore this pull request and review instead model_04_chimerafit pull request

Issue #1239, opened 2017-10-02
Motioncor2 shifts plot (wish list made by B13 people-IV)


rmarabini:
What is the meaning of the motioncor2 shift plots (since it can not be shifts). I will (@rmarabini) look into it.

delarosatrevin:
@azazellochg, can you comment on this? My understanding is that the plot are the average shift in a frame in x and y...

azazellochg:
@delarosatrevin yes, you are right. Those are cumulative per-frames shifts in pixels, first frame shift is always set to 0,0

rmarabini:
Hi @delarosatrevin  and @azazellochg,

 I have plots in which the difference between last and first frames is 600 pixels. If you look to the original movie versus the corrected one you can see a small improvement but there is no way the difference between first and last image is 600 px. Furthermore, if you open the log files the shifts are around 1 pixel beetwen frames. There is something fishy here

azazellochg:
Hi @rmarabini, could you attach the log file here? I can have a look

rmarabini:
@azazellochg <https://github.com/azazellochg>, give me a few days to
analyze the problem properly and I will send you the log files and the
plots.

  roberto

delarosatrevin:
@rmarabini , any news on this? There is a user that gave some feedback about the plots to @cossorzano , @pconesa and @josuegbl. It would be good to take a look into this.

rmarabini:
Hi @delarosatrevin,

I am working on it.

   Roberto

azazellochg:
I guess this is solved now.

PR #1356, opened 2017-11-22
Rm motioncor2

rmarabini:
Title:Fix motioncor2 plots

Symptoms: Motioncor2 plot shows shifts in the range of hundreds pixels. This result cannot be correct

Fix: shift X and Y are given with respect to the first frame, no accumulation -as we used to do in the protocol- is needed. See Motioncor2 user manual(http://msg.ucsf.edu/MotionCor2/MotionCor2-UserManual-10-19-2016.pdf): "The output and log files list the shifts relative to the first frame."

NOTE: I also double-checked that units are pixels. This seems to be true since since samplingrate is only asked by the program if dose filtering is required

NOTE: The protocol used to multiply shifts by the value of the "binarization factor" this is not needed. Motioncor2 report shifts in the "unbinned" pixel size. This information is not in the user guide but has been obtained by aligned the same data set with two values of binning

Warning: the program plots the content of the file: micrographs_0000X-Patch-Full.log. These are Initial alignment based upon full frames. These initial values are further refined using   a "Patch based alignment". The second refinement is in file micrographs_0000X-Patch-Frame.log (values seems to be 1-2 pixels vs about 6-8 typical difference between first and last frame). The avg shifts may be added to the previous ones.  Parsing the file micrographs_0000X-Patch-Frame.log is not difficult but requires some work. Any volunteer is welcome to do so.

Note: in magnitude, the shifts in Patch-Full.log are larger than the shifts stored in Patch-Frame.log so the plot as it is now is useful.

Comment:  binning factor in motioncor2 can only be 1 and 2. If somebody is going to modify the protocol wrapper we may change the binning factor text box by a radio button   

   Roberto

P.S: I add here the names of the programmers that appear in the protocol file header @delarosatrevin , @josuegbl, @azazellochg


azazellochg:
Thanks @rmarabini ! A few comments:
1) binning factor can be non-integer in motioncor2 (1.5)
2) first frame shifts are no longer 0,0 (in patch-full) in motioncor2 1.0.2 version
3) patch-frame file shows refined shifts for the patches after global frame shifts were applied, so they do not reflect real global frame drift. I would argue against using them or adding to global shifts. Moreover, it looks like all log files except patch-full still print 0,0 for the first frame (what a mess...).

pconesa:
@azazellochg , I'll deal with labeling conficts. Thanks.

pconesa:
Closes #1239

pconesa:
@azazellochg we have removed the `if meanX[0] != 0 or meanY[0] != 0`..what do you think. Also, we have checked that 1.0.2 shifts, seems to refer to the middle frame.

pconesa:
I've updated the [lastHTMLRepo](http://scipion.cnb.csic.es/scipionbox/lastHTMLReport/) and have this changes about the shift plot.

azazellochg:
Looks great so far, I have updated unblur protocol the same way. Can someone check optical flow/correlation protocols, so that plot would look the same across all movie processing programs?

azazellochg:
Hm, @pconesa it looks like you have not incorporated changes from facilities-fixes.. Like angstrom signs, 0.2f values etc...

pconesa:
umm. @azazellochg , you are right...I think all changes are here...last move was to move the units to the column headers...I guess this has "cause" the digits to appear. I'll have a look

PR #1364, opened 2017-11-24
Archive

kevinpetersavage:
This adds a new (recently created) protocol that can send messages to diamonds archive queue service. It is also a simple test of me breaking off small parts from release-1.1-headless-devel and checking that I can make pull requests based on this.


pyworkflow/em/packages/archiving/protocol_archive.py
```diff
@@ -0,0 +1,58 @@
...
+                       help='The host name for a queue broker (e.g. activemq)')
+        group.addParam('queuePort', params.IntParam,
+                       label="Port", important=True, default=61613,
+                       help='The stomp port for a queue broker (e.g. activemq)')
+        group.addParam('queueName', params.StringParam,
+                       label='Queue Name',
+                       help='The name of the queue within the broker')
+
+    def processMovieStep(self, movieDict, hasAlignment):
+        with open('/scratch/scipion/pyworkflow/em/packages/archiving/recipe.json') as recipe:
```
pconesa: Is this recipe.json, the same in the same path of this python file.
This could be not hardcoded like?
```
SCIPION_HOME = os.environ['SCIPION_HOME']
LOGFILE = os.path.join(SCIPION_HOME, 'pyworkflow', 'em', 'packages','archiving','recipe.json')
```

delarosatrevin: Yes, I agree with @pconesa that this json file would be better inside the package as a 'resource' of the package. Moreover, since this archiving is something related to Diamond, should not make more sense to put inside the ispyb package? Or even better, create a 'diamond' (or 'dls') package and have inside ispyb and achiving protocols? I guess 'ispyb' would not be a good package name if the people from ESRF starting adding other ispyb tools.

pyworkflow/em/packages/archiving/recipe.json
```diff
@@ -0,0 +1,23 @@
...
+         "parameters": { "files": ["{filePaths}"] },
+         "output": { "finally": 2 }
+       },
+  "2": { "service": "DLS Archiver",
+         "queue": "archive.pattern",
+         "transport-delay": 300,
+         "parameters": { "files": ["{filePaths}"],
+                         "limit-files": 1000,
+                         "dropfile": "{dropfile_override}",
+                         "dropfile-dir": "/dls/tmp/kevin/",
```
delarosatrevin: There are also some 'hard-coded' path in the .json file. If there is a configuration file I guess it should be better to be generated from a templated and not inside the source root.

kevinpetersavage:
I've addressed the hard coded paths, but I'm a little unsure what to do with the packages. Shall I just rename archiving to dls with the intent of moving ispyb later?

PR #1365, opened 2017-11-27
Don't use realpath when searching for mpi.

akesandgren:
OpenMPIs mpicc is a softlink to orted and the rest of the code will not
find the correct mpiHome.

### DIFF
--- a/pyworkflow/apps/pw_config.py
+++ b/pyworkflow/apps/pw_config.py
@@ -433,7 +433,7 @@ def guessMPI():
     for d in os.environ.get('PATH', '').split(':'):
         if not os.path.isdir(d) or 'mpicc' not in os.listdir(d):
             continue
-        mpiBin = os.path.realpath(join(d, 'mpicc'))
+        mpiBin = join(d, 'mpicc')
         if 'MPI_BINDIR' not in options:
             options['MPI_BINDIR'] = os.path.dirname(mpiBin)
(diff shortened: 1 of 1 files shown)


pconesa:
Thanks @akesandgren , we will have a look.
@delarosatrevin , could you please check this will not break other cases. Not sure why we bother to use realpath...

delarosatrevin:
Hi @pconesa , we used realpath because in some cases both mpicc and javac are symbolic link, so getting the realpath is a workaround to find MPI_HOME and JAVA_HOME. This is only for trying to figure out where are installed these libraries and it is useful for a general user that is not a sysadmin.

@akesandgren opinion is that using realpath is wrong for detecting MPI installation folder.

I think the more general solution is that we should not try to 'guess' where this libraries are and allow a more standard way to specify them depending on the system and check that they are properly configured.

akesandgren:
A late comment, you should never use MPI_HOME. Use the mpi compiler wrappers, they know where the includes and libs are.

pconesa:
Thanks I'll try to learn about this

pconesa:
Dear @akesandgren , @delarosatrevin this is being here for a long time. The reason is basically that there are 2 cases, one for which using the real path is ok and the other where not. Since, this is something that can be easily solved editing the config file. I'm closing this PR. In any case, thank you @akesandgren for this. We will still give a though about this.

PR #1396, opened 2017-12-18
Fix wrong input set size when working in streaming

### LINKED ISSUE #1310: Input size does not refresh while streaming.
The input sets do not show the actual set size.

pconesa:
Closes #1310


pconesa:
I think this is not a risky fix, but would like to have your OK, at this moment, just before releasing.

pconesa:
Wait, I have some refactor to do and remove a print line.

PR #1415, opened 2018-01-02
Web services not sockets

kevinpetersavage:
Happy new year!

This is a minor change from using raw sockets to using http which ends up handling opening and closing socket more reliably. 

I'm not quite sure why Github can't merge this, if I try the merge here it works.

PR #1416, opened 2018-01-02
eBic startup script

kevinpetersavage:
The start-up script used at eBic/Diamond. 

Note that there is a change to socket based file importing that was part of one of the original commits for some reason but shouldn't be needed/make any difference so I tried not to separate it out. I can do this if you'd like me to.


pyworkflow/em/protocol/protocol_import/micrographs.py
```diff
@@ -485,11 +485,16 @@ def iterFilenamesFromSocket(self):
                         self.debug("Data received in socket:")
                         self.debug(files)
                         for fileName in files:
-                            uniqueFn = self._getUniqueFileName(fileName, files)
-                            if uniqueFn in self.importedFiles:
-                                self._spreadMessage('WARNING: Not importing,'
-                                                    ' already imported file %s '
-                                                    '\n' % fileName, sock)
+                            if os.path.exists(fileName):
```
josuegbl: I think that the previous codes were correct and should not modified. There is any reason to change it? Seems to be a wrong merge, that its correct?

kevinpetersavage: Yes, I think that was part of a commit that was mainly about a different file, sorry. 

It doesn't actually matter because this other pull request https://github.com/I2PC/scipion/pull/1415#issuecomment-354811817 removes that section completely. On that basis would you like be to revert this change from this pr or will it be easier to merge if I don't?

pconesa:
@yaizar , shouldn't this PR be just the script?

yaizar:
I don’t know, Do you mean the mixed read-only script? That one is on #1471

pconesa:
@yaizar , my bad I was confused with #1471 PR.

pconesa:
At some point, we should catch up with DLS to update their setup which should be easy now since their are back to normal import using glob.

PR #1419, opened 2018-01-09
Release 1.1.facilities fixes

josuegbl:


pyworkflow/em/packages/grigoriefflab/protocol_ctffind.py
```diff
@@ -258,7 +258,7 @@ def _prepareCommand(self):
         # Convert digital frequencies to spatial frequencies
         self._params['sampling'] = sampling
         self._params['lowRes'] = sampling / self._params['lowRes']
-        if self._params['lowRes'] > 50:
+        if self._params['lowRes'] > 50 and self.useCtffind4:
```
azazellochg: A value for Resmin of anything lower (e.g. 100A) will be reset to 50 A by ctffind anyway, so there is no point...

pconesa: @josuegbl mentioned that at some point he was getting errors? @josuegbl ?

josuegbl: ctffind4 fails if lowRes is lower than 50A, but ctffind3 not. Ctffind3 sometimes is used with values lower than 50A, especially with negative stain micrographs.

azazellochg: sure, you can pass values lower than 50A to ctffind3, but it will not change anything:
```
...
      STEPR=DSTEP*(10.0**4.0)/XMAG
      RESMIN=STEPR/RESMIN
      RESMAX=STEPR/RESMAX
      IF (RESMIN.LT.STEPR/50.0) THEN
        RESMIN=STEPR/50.0
        WRITE(*,*)' Lower resolution limit reset to',
     +            STEPR/RESMIN,' A'
      ENDIF
```

josuegbl: Thanks a lot @azazellochg, I didn't know it and more than one "expert user" doesn't know it either. I'll restore the code.

PR #1420, opened 2018-01-09
Generates an output with the manually discarted coordinates

dmaluenda:
When the Xmipp protocol automatically picks a set of particles, we can discard some outliers to refine the autopicking procedure. Now, this outliers are also saved in a separated coordinates output to be used in other protocols (a form's flag can revert this new feature).


josuegbl:
@dmaluenda, please add a test to this new feature and check if everything works ok, focusing protocols that uses the functions that you have been modified. Also, test it in streaming.

PR #1439, opened 2018-01-25
Use pip for python modules

pconesa:
All python modules installed through pip, except Scons
Scons not done because it goes to /bin.
SConstruct and all installations use the SCIPION_SOFTWARE variable.
Path for lib, include and bin centralized in Environment

Python updated to 2.7.14
Sqlite3 updated to 3.21
Installation script should clean previous python and sqlite installation if found.


SConstruct
```diff
@@ -215,7 +210,7 @@ def addCppLibrary(env, name, dirs=[], tars=[], untarTargets=['configure'], patte
     targetName = join(basedir, target if target else prefix + name)
     sources = []
 
-    _libpath.append(Dir('#software/lib').abspath)
+    _libpath.append(Dir(SCIPION_SW + '/lib').abspath)
```
delarosatrevin: From my old Scons memory I think that the prefix "#" in the path string meant the Scons building directory, but not sure.

install/funcs.py
```diff
@@ -254,14 +254,47 @@ def getLibSuffix(self):
     def getProcessors(self):
         return self._processors
     
+    @staticmethod
+    def getSoftware():
```
delarosatrevin: This method could be more implemented more compact:

return os.environ.get('SCIPION_SOFTWARE', 'software')

SConstruct
pconesa: Thanks @delarosatrevin , actually I think it means "relative" but I'm not sure. SCIPION_SW comes as an absolute path like "/home/user1/...." and "#" + SCIPION_SW was turn into "home/user1/..." (note there is no starting /) and ended up being <SCIPION_HOME>/home/user1/...".

install/funcs.py
pconesa: ahh!  one day I'll be 100% pythonic! ;-) thanks.

SConstruct
delarosatrevin: Yes, it could be...I only wanted to warn about the special meaning of #

PR #1449, opened 2018-02-02
Aj movie resize

ajimoreno:
Added a protocol to make a resize of SetOfMovies (able to manage input in streaming) (protocol_movie_resize.py). Added the corresponding test (tests_protocols_xmipp_movie_resize.py).
Moreover, the protocol create stream data has been modified to receive a SetOfMovies as input, instead of a Movie.


pyworkflow/em/packages/xmipp3/protocol_preprocess/protocol_movie_resize.py
```diff
@@ -0,0 +1,196 @@
...
+from pyworkflow.em import STEPS_PARALLEL
+from pyworkflow.em.data import SetOfMovies, Movie
+from os.path import exists
+
+class XmippProtMovieResize(ProtProcessMovies):
+    """
+    Resize a set of movies.
+    """
+    _label = 'movie resize'
+    _lastUpdateVersion = VERSION_1_1
```
pconesa: I think this has to be VERSION_1_2 at least since it is new.

Issue #1402, opened 2017-12-21
Default value for 'Use previous alignments' in Relion


delarosatrevin:
From @JuhaHuiskonen...

I noticed that 'consider previous alignment' is a default in relion refinement protocols.   If this is selected, the current angles get written as priors, which is what it should do. But I don't think it makes sense to have this as a default however. 
 
I am saying this because I think it causes a problem in most typical workflows, where the angles come from 3D classification with a very coarse sampling. E.g. the angels can be off by let's say 7.5 degrees. In auto-refine, they get first refine in 'global mode'. But when the refinement switches to 'local mode', the local mode focuses on the Prior angle, not the best angle from the current refinement.  

In other words, having 'consider previous alignment' makes only sense, when the user wants to skip global alignment and switch directly to local alignment, for example in focused classification (or localised reconstruction).  

In summary, I would not have this as a default. And perhaps describe the behaviour (that the old angles are considered as Priors) in the help. 

@azazellochg , @josuegbl  Can you comment on this? I agree with Juha's point. 
@pconesa I think if we agree on the change, it should be very easy to implement (just put False as default value in that option), so it could enter in the release.

pconesa:
Thanks, if all both agree, yes it can go in the release. By the way, I don't see it happening this year!.

azazellochg:
Hi @delarosatrevin ! I agree with @JuhaHuiskonen , the current default option will set priors and thus all subsequent local searches are done around those. Switching this off will allow optimal angles/shifts (from previous iteration) to slowly move away.

However, now the alignments are not copied from input particles at all to input star file if this options is False. If someone whats to continue from previous iteration (but not stick to priors), continue option should be used with copyAlignment=False. But what if I want to do 3D classification using angles from 3D refinement without sticking to priors? Continue cannot be used now, because it is a different protocol.

pconesa:
Thanks @azazellochg . @JuhaHuiskonen , @delarosatrevin ... Is this becoming a not as straightforward as initially thought?

JuhaHuiskonen:
Hi @azazellochg @pconesa @delarosatrevin so should there by two separate options:
1. consider previous alignment (not a default - if selected will copy the alignments but not the priors)
2. consider alignment as priors (not a default and perhaps this should be an advanced option evoked only if 1 is chosen - if selected will copy the alignments also as priors).

josuegbl:
[...]

PR #1454, opened 2018-02-05
Release 1.1.facilities fixes

delarosatrevin:
### Summary of the changes in this PR:
* Relion:
  * Changed how the Magnification is handled in Relion convert.py 
  * Some fixes in the bibtex
  * Changes in how input Alignment is treated (plus added priors)
  * Bug fixed in protocol base using 'realignMovieFrames' that is not defined in classify2D
  * Bug fixed in extract-particles when extracting from a subset
* Ctffind4 protocol now allow to set Yes/No for slow/fast search 
* In scipion script the 'v' of the version name is used in lowercase for more consistent naming
* Some values persisted in screen-particles protocol to be used in summary and test validation.
* Add radius size validation in extract movie particles protocol

### DIFF
--- a/pyworkflow/em/packages/relion/convert.py
+++ b/pyworkflow/em/packages/relion/convert.py
@@ -360,7 +360,7 @@ def alignmentToRow(alignment, alignmentRow, alignType):
         angle = angles[0] + angles[2]
         alignmentRow.setValue(md.RLN_ORIENT_PSI, -angle)
         # Also set the prior
-        alignmentRow.setValue(md.RLN_ORIENT_ROT_PRIOR, -angle)
+        alignmentRow.setValue(md.RLN_ORIENT_PSI_PRIOR, -angle)
 
         flip = bool(numpy.linalg.det(matrix[0:2,0:2]) < 0)
         if flip:
--- a/pyworkflow/em/packages/relion/convert.py
+++ b/pyworkflow/em/packages/relion/convert.py
@@ -160,7 +160,7 @@ def isVersion2():
 
 
 def getSupportedVersions():
-    return [V1_3, V1_4, V2_0, V2_1]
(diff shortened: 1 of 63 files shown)


pyworkflow/em/packages/relion/convert.py
```diff
@@ -359,6 +359,9 @@ def alignmentToRow(alignment, alignmentRow, alignType):
     if is2D:
         angle = angles[0] + angles[2]
         alignmentRow.setValue(md.RLN_ORIENT_PSI, -angle)
+        # Also set the prior
+        alignmentRow.setValue(md.RLN_ORIENT_ROT_PRIOR, -angle)
```
azazellochg: should this be md.RLN_ORIENT_PSI_PRIOR ?

azazellochg:
Would it be reasonable address #1402 here? Or you want to postpone it?

delarosatrevin:
Yes, I think it should be addressed here. I thought that this was taken into account by @josuegbl in the changes related to Relion. @josuegbl Can you double-check?

pconesa:
@delarosatrevin and @josuegbl ....I've pushed my fixes for the screen particle. I didn't like the summaryVars thing since then is "very dirty" to test that those values are right, even if only used in the summary. If scipion is used as an API....this way, these values could be easily retrieved. Hope you don't mind to leave it as it is.

pyworkflow/em/packages/relion/convert.py
josuegbl: you are right @azazellochg, it was an error introducing the label :)

delarosatrevin:
@josuegbl Can you take a look at this PR and merge before it continues to grow? Some other minor changes has been introduced since @pconesa approved it. ;)

Issue #1083, opened 2017-06-28
Have multiple parallel_command in host


pconesa:
From Grigory:

Provide a different PARALLEL_COMMAND in hosts.conf for a specific queue.

Would be also great to have multiple SUBMIT_TEMPLATE...

PR #1455, opened 2018-02-05
Allow to specify different hosts.conf through an environment variable

delarosatrevin:
In this PR two things are accomplished:
* Replace the use of the string 'hosts.conf'  by the variable SCIPION_CONFIG_HOSTS and,
* Allow that value to set externally via a environment variable. (This is useful to reuse the same installation for workstations and cluster which requires different hosts.conf files)

NOTE: I have created this branch after merging with facilities-fixes, so that PR should be merged first in order to see real changes introduced here.


pconesa:
@delarosatrevin , isn't it the other way around. We must accept first facilities fixes?

delarosatrevin:
Yes, that was what I intended to say, but you know...using a 'this' instead of a 'that' can mess it all over. I changed the original message. So, summarizing...facilities-fixes first and then this one.

azazellochg:
I guess this "kind of" solves #1083 :)

delarosatrevin:
@azazellochg Not sure if exactly the same...I think you want to have this in the same hosts.conf and in the same machine...and the difference should be by queue...

pconesa:
@azazellochg , thanks.

PR #1470, opened 2018-02-12
Release 1.1.facilities fixes

josuegbl:
added new param alignmentAsPriors to consider the priors in relion metadata.
added motioncor2 1.0.5


pyworkflow/em/packages/grigoriefflab/grigoriefflab.py
```diff
@@ -27,6 +27,12 @@
 import os
 from os.path import join, exists
 
+# Program common names
+FREALIGN_APP = 'FREALIGN'
```
delarosatrevin: Why not just FREALIGN and CTFFIND4?

delarosatrevin:
@josuegbl I will review this PR, but first I will like to add some missing parameters (related to number of subsets and subset size) to Relion 2D and 3D classification.

pyworkflow/em/packages/grigoriefflab/grigoriefflab.py
pconesa: Yep, there is already a constant with that name

delarosatrevin: With a different value/purpose?

pconesa: Yes! I thought the same, but found that other variable and didn't want to "pull the blanket" (;-))

delarosatrevin:
@josuegbl If the copy alignment as priors is used when the input particles has not alignment information (or this is my guess) it is failing in the convertInputStep, I think it should be validated before executing the protocol. If I launch the same protocol from particles with alignment it seems to work. (see image below)

image

josuegbl:
@delarosatrevin you are right! Ill fix it.

PR #1471, opened 2018-02-12
Import readonly project

yaizar:
Added script to import a read only project so that we can add more protocols to the workflow while keeping the initial runs as read only.


pconesa:
This is now old....DLS is moving to 1.2. We might need this changes there or to a new branch from 1.2.

PR #1481, opened 2018-02-14
gctf on particles

### LINKED ISSUE #912: ctfModel for particles
A while ago I was thinking to add local ctf refinement at particle level (by Gctf). I was wondering how much modification of CTFModel class will be required, also to keep backward compatibility?

azazellochg:
This PR adds a new gctf refine protocol to refine CTF locally on particles. There are several things to note:

- Gctf estimates CTF de-novo for the full mic, and then refines locally for each particle. This is default behaviour.
- This means the separate protocol is almost the same as original one, but input particles and mics are required. I didn't find a way to get mics' sampling and full filename from particles set only.
- Right now particle set is matched with output from Gctf by looking at micname + coordinates.
- The protocol works sequentially on each micrograph.
- No streaming implemented..

This PR closes #912

TODO:

- [x] add test protocol
- [x] remove re-estimate CTF options, since re-estimation is probably not needed here?
- [x] add refinement options of input CTFs (use as starting point for local refinement)


delarosatrevin:
Thanks a lot @azazellochg. I think this is an important feature from Gctf that some users had requested. I will review this PR. I think that having a test could be useful (not only for testing now, but also for the future updates).

azazellochg:
Test files to upload:
- BPV_1386.stk, BPV_1387.stk, BPV_1388.stk, BPV_particles.sqlite, BPV_particles_aligned.sqlite into data/tests/xmipp_tutorial/particles/
- ctfs.sqlite into data/tests/xmipp_tutorial/ctfs/

https://drive.google.com/open?id=1LI1qqXW02d3eCwXlEP8HSJkUI-CsmHwr

pconesa:
Files added!

delarosatrevin:
Thanks @pconesa ! I'm "just" recompiling this branch (devel based) to review this PR. ;)

delarosatrevin:
@azazellochg I made some modification and test are still working. Can you double-check that I haven't broke anything and merge the PR?

By the way, I'm also picking these changes to my branch based on v1.2 (sdevel and smaster)

PR #1488, opened 2018-02-19
Devel fixes

cossorzano:


pyworkflow/em/packages/eman2/protocol_autopick.py
```diff
@@ -66,6 +68,7 @@ def _defineParams(self, form):
         form.addParam('gaussWidth', FloatParam, default='1',
                       label='Gauss Width',
                       help='Width of the Gaussian kernel used')
+        form.addParallelSection(threads=1, mpi=4)
```
josuegbl: I dont know why is necessary run this protocol in parallel, since is one of the fastest pickings inside Scipion. In general, this protocol takes less than 5 sec per micrograph. So, where is the need to parallelize? This changes has been probed?

Issue #1493, opened 2018-02-20
Join set fails


pconesa:
It is failing in a join of sets of particles. Apparently, the attributes are the same...but need to scan deeper in the attributes because de _ctfmodel differs. One has an extra parameter

_ctffind4_ctfPhaseShift

I've done some work. @rmarabini check with me in case you've done something to fix this.

PR #1496, opened 2018-02-21
Allow join sets of different attributtes getting only the common ones…

pconesa:
…. Warn about it.

### DIFF
--- a/pyworkflow/object.py
+++ b/pyworkflow/object.py
@@ -129,12 +129,30 @@ class Object(object):
         subclasses of Object and will be stored"""
         for key, attr in self.getAttributes():
             if not hasattr(attr, '_objDoStore'):
-                print "Object.getAttributesToStore: attribute '%s' seems to be overwritten," % key
-                print "   since '_objDoStore' was not found. Ignoring attribute. "
+                print ("Object.getAttributesToStore: attribute '%s' seems to "
+                      "be overwritten," % key)
+                print ("   since '_objDoStore' was not found. "
+                       "Ignoring attribute. ")
             else:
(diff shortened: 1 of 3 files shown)


pconesa:
Closes #1493

pyworkflow/object.py
```diff
@@ -129,12 +129,30 @@ def getAttributesToStore(self):
...
+                print ("Object.getAttributesToStore: attribute '%s' seems to "
+                      "be overwritten," % key)
+                print ("   since '_objDoStore' was not found. "
+                       "Ignoring attribute. ")
             else:
                 if attr is not None and attr._objDoStore:
                     yield (key, attr)
-            
+
+    def getAttributesToStoreEx(self, prefix=""):
```
delarosatrevin: @pconesa I think we don't need an extra function in Object. Have you checked Object.getObjDict? the keys of this dictionary are (almost) the strings returned by this new function. I think will be better to reuse the getObjDict function and avoid to add additional ones that are used only in very very specific case.

pconesa: I'm looking at it...Actually, I didn't put it there initially...only after @josuegbl suggestion I move it there.

delarosatrevin: My point is that maybe the function is not needed at all.

pconesa: I think it returns exactly what I need...thanks @delarosatrevin

pconesa:
@delarosatrevin , function removed.

delarosatrevin:
Thanks a lot! I will continue reviewing this PR

PR #1497, opened 2018-02-22
Release 1.1.facilities fixes

### LINKED ISSUE #1494: Xmipp Particle Crop fails
Cropping particles fail when converting CTF model to xmipp row.
It is failing because the ctf has a PhaseShift set to null. ShowJ change it to 0.000 but xmipp convert does not do that and fails:

> ERROR:  Protocol failed: float() argument must be a string or a number

pconesa:
Manage phaseShift:
Relion convert phaseShift from CTFModel and back
Test the conversion
Xmpp does not use phaseShift

Closes #1494 

Sorry to bother guys, but since we are close to the release...I would appreciate your review on this. 8 eyes are better than my poor 2.

### DIFF
--- a/pyworkflow/em/data.py
+++ b/pyworkflow/em/data.py
@@ -31,14 +31,10 @@ for EM data objects like: Image, SetOfImage and others
 
 import os
 import json
-
 from pyworkflow.object import *
-import pyworkflow.utils as pwutils
-
 from constants import *
 from convert import ImageHandler
 import numpy as np
-# import xmipp
 
 
 class EMObject(OrderedObject):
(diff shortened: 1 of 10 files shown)


pyworkflow/em/data.py
```diff
@@ -125,7 +125,7 @@ def __init__(self, **kwargs):
         self._defocusV = Float(kwargs.get('defocusV', None))
         self._defocusAngle = Float(kwargs.get('defocusAngle', None))
         self._defocusRatio = Float()
-        self._phaseShift = Float()
+        self._phaseShift = Float(kwargs.get('phaseShift', None))
```
josuegbl: I think is better something like: self._phaseShift = None and not store in db its value. If you set self._phaseShift = Float(kwargs.get('phaseShift', None)), you are storing to None in the db and this could be problematic and time wasting in the case when you need to convert millions of particles....

delarosatrevin: I agree.

pconesa: I'm changing this.

pconesa: Need to push some my changes to test then in a GPU machine....wait for my Green light to review/accept.

pconesa:
Thanks Gregory for testing, my changes will imply that CTFModel may or may not have phaseShift so I'm updating the converts so those packages...bear with me until I finish my changes.

pconesa:
Ok, I've tested the five ctf estimations combinations (ctffind4 w and w/o phaseShift), gCTF (x2), and xmipp (x1)

I've extracted particles from all of them using relion and xmipp and checked whether phaseShift is present or not. For me it all looks correct.

josuegbl:
I do not like that the treatment given to phaseShift in our model is totally different than the other ctf parameters. I think more homogenization is better for all of us.

delarosatrevin:
Hum...here I'm a bit divided between @josuegbl concerns and @pconesa solution. I agree with @josuegbl about some differences in the treatment of this parameters and other in CTFModel...so I will propose to make it more homogeneous. On the other side, I understand some of the hacks to reach a solution right now. 

@pconesa Should I try to doing some changes regarding this and we can discuss them after that?

josuegbl:
Im changing the code. We can discuss when submit the changes.

josuegbl:
@delarosatrevin review this and merge if it is ok for you ;)

PR #1503, opened 2018-02-27
Protocol to trigger data

majtner:
- added protocol to trigger data (for streaming purposes)


pyworkflow/em/packages/xmipp3/protocol_trigger_data.py
```diff
@@ -0,0 +1,207 @@
...
+                           'particles will be send to output.')
+        form.addParam('delay', IntParam, default=1,
+                      label="Delay (sec)",
+                      help="Delay in seconds before checking new output")
+
+    # --------------------------- INSERT steps functions ----------------------
+    def _insertAllSteps(self):
+        self.finished = False
+        self.AddedParticles = []
+        partsSteps = self.delayStep()
```
pconesa: This is not actually inserting any step.
You are just doing all this without steps?

majtner:
@pconesa Thank you for comments. 
I have changed the names of variables and used some code from protocol_movies to optimize _checkNewInput (I am not sure if this is the one you had in mind, probably not but it will not harm there). I was not able to find "creation time" of single particle, there is no method for it, so I can wait for your help, if there is even better way to process checkNewInput.

About the execution: I am adding only delayStep, because checkNewInput and checkNewOutput are called automatically and there is no "processing" for creating additional steps. The protocol is only waiting for enough inputs to pass them to output.

cossorzano:
Should we add a test?

majtner:
Ok, I am still making some changes and I can add also a test.

majtner:
- in _checkNewInput, there is new condition for creation time (based on discussion with @pconesa)
- update of create_stream_data protocol for streaming particles
- added test for trigger data protocol

pyworkflow/em/packages/xmipp3/protocol_trigger_data.py
```diff
@@ -0,0 +1,221 @@
...
+class XmippProtTriggerData(ProtProcessParticles):
+    """ Waits until certain number of particles is prepared and then
+    send them to output. """
+
+    _label = 'trigger data'
+
+    # --------------------------- DEFINE param functions ----------------------
+    def _defineProcessParams(self, form):
+
+        form.addParam('outputSize', IntParam, default=60,
```
pconesa: @majtner , is 60 here a reasonable default for outputSize?

majtner:
@pconesa Default value was changed, thank you.

majtner:
@ajimoreno In case you are interested in trigger data protocol, it was already merged with devel, so you can use it.

PR #1506, opened 2018-03-01
Allow Set's iterators to reduce it's scope with a limit and skipRows

pconesa:
Optional behavior

Set and mappers accept new optional parameters
Tests for flat mapper and sets.

### DIFF
--- a/pyworkflow/mapper/sqlite.py
+++ b/pyworkflow/mapper/sqlite.py
@@ -866,7 +866,9 @@ class SqliteFlatMapper(Mapper):
                       , objectFilter=None
                       , orderBy=ID
                       , direction='ASC'
-                      , where='1'):
+                      , where='1'
+                      , limit=None
+                      , skipRows=None):
         # Just a sanity check for emtpy sets, that doesn't contains 'Properties' table
         if not self.db.hasTable('Properties'):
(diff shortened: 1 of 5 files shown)


pyworkflow/mapper/sqlite.py
```diff
@@ -1162,6 +1165,20 @@ def _getRealCol(colName):
...
+        # If there is a limit
+        if limit:
+            # if it is a tuple
+            if isinstance(limit, tuple):
+
+                skipRows = limit[1]
+                limit = limit[0]
+
+            # If we need to skip rows
+            skipPart = "%s," % skipRows if skipRows else ""
```
delarosatrevin: This code fails if limit is not a tuple because skipRows is not defined.

delarosatrevin: I just added a PR to this branch with the fix.

pconesa:
thanks @delarosatrevin , I forgot about this. @yaizar will add here another fix in the pip installation that fixes the issue you had installing pip: basically when pip is installed locally for the user.

delarosatrevin:
Great @pconesa!. I think it is fine from my side...so you can merge it after these changes.

PR #1511, opened 2018-03-02
Model 05 refmac

mmmtnez:
Dear all,

As we said in our five previous pull requests, we divided the large model_building pull request in some smaller ones. This sixth pull request (model_05_refmac), coming after the coot refinement (mm_modeling), deals with the automatic refinement of an atomic structure regarding a density volume. 
model_05_refmac contains the update of mm_modeling, which included the rigid fitting of atomic structures to density volumes (with powerfit and chimera) and the manual fitting/refinement program coot. In addition, model_05_refmac includes the refmac protocol to carry out the automatic refinement.

All the protocols make intensive use of the "origin" added to volumes and consider the possibility of "origin" for atomic structures.

In detail:
1.- We have included in all the protocols, including the pseudoatoms protocol, the origin coordinates of the volume. In case these coordinates exist, they will be loaded. Otherwise, the default coordinates will be inserted.
2.- We have modified the protocol to import volumes (volumes.py) to give the user the possibility of choosing their own origin coordinates or, instead, using the default origin coordinates (that place the volume in the centre of the coordinate axes). If he decides to introduce their own origin coordinates, first he has to select the option YES in the form param "Set origin of coordinates", and then write directly the coordinates inside x, y, and z boxes that will appear. Next to those boxes, a wizard suggests the user to insert the origin coordinates present in the map header (if they exist there).
3.- We have modified the relationships between inputs and outputs in powerfit and chimera fit/chimera operate protocols, associating the input volume to the output volume and the input pdb to the output pdb. In case there is not input volume, the output volume will be related with the input pdb.
4.- Previously we had the same form for chimera operate and chimera fit protocols. In this update, we have set particular forms for each protocol. Particularly, we have modified the label and the help of pdbFileToBeRefined and inputPdbFiles params that, now, will be different for chimera operate and chimera fit protocol forms.
5.- The new protocol included, protocol_refmac, displays results in a table so that the user could select the particular result that he wants to see. In the first graphical display, the input pdb (usually the coot protocol output), the input reference volume, the masked map obtained from the first pdb generated by refmac (in the first run), and the final refined pdb will be shown.

Test: scipion test tests.em.protocols.test_protocols_import_volumes.TestImportVolumes
Test: scipion test em.packages.xmipp3.tests.test_protocol_extract_unit_cell.TestProtModelBuilding
Test: scipion test em.packages.chimera.test_protocol_chimera_operate.TestChimeraOperate
Test: scipion test tests.em.workflows.test_workflow_modeling.TestPowerFit
Test: scipion test tests.em.workflows.test_workflow_modeling.TestChimeraFit
Test: scipion test tests.em.workflows.test_workflow_modeling.TestCootRefinement
Test: scipion test tests.em.workflows.test_workflow_modeling.TestRefmacRefinement


pconesa:
@mmmtnez  did you merged it yourself? i'm confused.

PR #1520, opened 2018-03-06
Er autosharpen

erneyramirez:
Phenix package has been integrated in scipion. The protocol autosharpen has been implemented.


pyworkflow/em/data.py
```diff
@@ -729,10 +729,10 @@ def hasClassId(self):
         return self._classId.hasValue()
 
     def hasHalfMaps(self):
-        return self._halfMapFilenames.hasValue()
+        return not self._halfMapFilenames.isEmpty()
 
     def getHalfMaps(self):
-        return self._halfMapFilenames.get()
+        return self._halfMapFilenames
```
delarosatrevin: Hi @erneyramirez This change now returns a different type...before was a Python string or None...now I think will be our String object....they are similar but not the same...why this change? Have you checked that other places in the code where this function is used?

pyworkflow/em/data.py
```diff
@@ -729,10 +729,10 @@ def hasClassId(self):
         return self._classId.hasValue()
 
     def hasHalfMaps(self):
-        return self._halfMapFilenames.hasValue()
+        return not self._halfMapFilenames.isEmpty()
```
delarosatrevin: I guess this changes wants to consider empty string as not having half maps, would be good to check that usages of the function are not affected.

pyworkflow/em/data.py
pconesa: @erneyramirez , changes here seems dangerous...I agree with @delarosatrevin. This changes must be thorougly tested and justified, unless the old code is proven to be failing.

pyworkflow/em/packages/phenix/viewer.py
```diff
@@ -0,0 +1,50 @@
+# **************************************************************************
+# *
+# * Authors:     Carlos Oscar Sorzano (coss@cnb.csic.es)
```
pconesa: Should @erneyramirez also appear here?

pyworkflow/em/data.py
pconesa: I think I know this.....this actually wasn't working since data is CSV and the hasValue does not work for this type as expected. And if I'm right, there wasn't any "consumer" of this method until now.

josuegbl: Without it serving as precedent, Pablo has right. :smile:

pyworkflow/em/data.py
josuegbl: This code was change by me because self._halfMapFilenames.get() returns a string with both map paths instead of a list. Also, this method is not used in anywhere.

pconesa:
I'm closing this. @erneyramirez will push updates soon and create a PR.

PR #1523, opened 2018-03-12
gpu classifier in no streaming and semi streaming modes

ajimoreno:
In this request I have included a new xmipp program to make 2D image alignment in GPU.
Two new protocols were created based on the previous program, "protocol_classification_gpuCorr" and "protocol_classification_gpuCorr_semi". The first one generates a set of 2D classes from a set of input images (no streaming), the second one is able to classify in streaming a set of images in a set of previously calculated classes (is called semi streaming because it needs some previous classes or averages). Their corresponding tests are also included in test_protocols_gpuCorr_classifier and test_protocols_gpuCorr_semiStreaming.
Also it must be highlight some other relevant changes: 
First, in SqliteFlatDb in the selectAll function, 'creation' has been included to be able to ask for objects with certain creation time values.
Second, in Protocol class, the _stepsCheckSecs label was included to be able to change this interval time in every protocol.
Several tests were executed to check that these changes do not break any other thing.


ajimoreno:
Solved.

Issue #1528, opened 2018-03-13
TestWorkflowRelionExtract fails


yaizar:
Assertion error [here](https://github.com/I2PC/scipion/blob/release-1.1.facilities-devel/pyworkflow/tests/em/workflows/test_workflow_relion.py#L183), defocusV expected is 24779 and it's hardcoded, value obtained is 24745:
![image](https://user-images.githubusercontent.com/4609687/37340693-32de61d6-26bf-11e8-8d8a-f46b56dda5b6.png)

PR #1536, opened 2018-03-19
Fixed bug in Relion extract - particles and PatchOverlap hidden when not MC2

delarosatrevin:
@josuegbl I think there was a condition introduced by you (img.hasCTF()) that was causing all rows in the star file had the same defocus values. I just removed it and now seems to work as expected, can you double-check the changes?

### DIFF
--- a/pyworkflow/em/packages/relion/protocol_extract_particles.py
+++ b/pyworkflow/em/packages/relion/protocol_extract_particles.py
@@ -561,10 +561,9 @@ class ProtRelionExtractParticles(em.ProtExtractParticles, ProtRelionBase):
         # The command will be launched from the working dir
         # so, let's make the micrograph path relative to that
         img.setFileName(os.path.join('extra', newName))
-        if self._useCTF() and not img.hasCTF():
-            img.setCTF(self.ctfDict[img.getMicName()])
 
         if self._useCTF():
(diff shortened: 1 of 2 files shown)


pconesa:
Josué is off not sure if he will be able to review this soon. I'll do my best also

pyworkflow/em/packages/relion/protocol_extract_particles.py
```diff
@@ -561,10 +561,9 @@ def _preprocessMicrographRow(self, img, imgRow):
         # The command will be launched from the working dir
         # so, let's make the micrograph path relative to that
         img.setFileName(os.path.join('extra', newName))
-        if self._useCTF() and not img.hasCTF():
```
pconesa: Should' we keep the check img.hasCTF() but without the "not"

delarosatrevin: This condition is to set the CTF to that img (micrograph), so it makes no sense to check img.hasCTF()

pconesa: I see, thanks

pconesa:
Closes #1528

pconesa:
I'm merging this.

PR #1539, opened 2018-03-21
Fixing problems with elimination of empty parts

majtner:
- changing condition in convert.py, which was not used before
- denoising is by default turned off (it is too slow with denoising for large sets of particles)
- optional denoising could be now balanced by parameter
- performance was enhanced by removing cloning particles
- c++ program stores all the good and bad particles in files, in streaming also increment/temp files are used to make process faster but later, these temp files are removed
- doneList in no longer used
- it was tested with streaming, without streaming, and without streaming on very large dataset, also TestXmippEliminatingEmptyParticles is ok

### DIFF
--- a/pyworkflow/em/packages/xmipp3/protocol_eliminate_empty_particles.py
+++ b/pyworkflow/em/packages/xmipp3/protocol_eliminate_empty_particles.py
@@ -29,9 +29,10 @@ import pyworkflow.em as em
 import pyworkflow.em.metadata as md
 import pyworkflow.protocol.constants as cons
 import pyworkflow.protocol.params as param
+from datetime import datetime
 from pyworkflow.em.protocol import ProtClassify2D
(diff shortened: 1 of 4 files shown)


pyworkflow/em/packages/xmipp3/protocol_eliminate_empty_particles.py
```diff
@@ -98,48 +109,46 @@ def _getFirstJoinStep(self):
                 return s
         return None
 
-    def _insertNewPartsSteps(self, insertedDict, inputParticles):
-        """ Insert steps to process new movies (from streaming)
-        Param:
-            insertedDict: contains already processed movies
-            inputParticles: input particles set to be checked
-        """
+    def _insertNewPartsSteps(self):
```
pconesa: @majtner , much cleaner!

PR #1541, opened 2018-03-23
Model 06

mmmtnez:
Hi all,
here you are our last small pull request to add the appropriate viewer to the chimera restore session protocol. This protocol, that you can find in SCIPION menu (Tools/Calculators) together with chimera operate protocol, allows restore chimera sessions that the user saves by default when he/she saves something in chimera operate and chimera fit protocols (or on purpose using the scipionss command in chimera command line). The viewer of the restore chimera session protocol will show the session saved in this protocol. In case nothing had been saved (so nor the session), the viewer will show the session of the parent protocol (chimera operate or chimera fit).

No tests have been included to show the functioning of this viewer because chimera session can only be saved with the graphical display, and tests run in --nogui mode. So, our tests of chimera operate and chimera fit do not save chimera sessions and it is impossible to restore them.


rmarabini:
This is MArta's only work so I feel comfortable reviewing it.

rmarabini:
Let us merge this

PR #1542, opened 2018-03-23
add relion extract movie ptcls

### LINKED ISSUE #1382: add relion extract movie particles

azazellochg:
This closes #1382 
In this PR the new protocol +test for relion 2.0+ is added. I could not find in test datasets any particles that would correspond to test movies (from the import movie test), so an upload of new particle set is needed for new test to work: `data/tests/relion_tutorial/import/refine3d_case2/*` 

PS. Release-1.2 could be merged into devel before this PR.


delarosatrevin:
@azazellochg I will review this PR as soon as I can and also take a look at the convertOutputStep.

pconesa:
I'm off.but will have a look next week

pyworkflow/em/packages/relion/convert.py
```diff
@@ -352,15 +353,10 @@ def alignmentToRow(alignment, alignmentRow, alignType):
 
     alignmentRow.setValue(md.RLN_ORIENT_ORIGIN_X, shifts[0])
     alignmentRow.setValue(md.RLN_ORIENT_ORIGIN_Y, shifts[1])
-    # Also set the priors
```
delarosatrevin: Are this changes intentional, or just the result of merging release-1.2 before into devel? The same for RLN_RANDOM_SUBSET...

pyworkflow/em/packages/relion/protocol_extract_particles_movies.py
```diff
@@ -0,0 +1,431 @@
...
+        movieParticles.setSamplingRate(self._getNewSampling())
+
+        mData = md.MetaData()
+        mdAll = md.MetaData()
+
+        for movie in self.inputMovieSet:
+            movieName = movie.getMicName()
+            # 'aligned_movie' suffix comes from protocol_align_movies
+            movieParts = movieName.replace('.mrcs',
+                                           '_aligned_movie_extract.star')
```
delarosatrevin: Is there any more general way to do this mapping? It seems a bit hard-coded to me.

pyworkflow/em/packages/relion/protocol_extract_particles_movies.py
```diff
@@ -0,0 +1,431 @@
...
+                mdAll.addItemId()
+
+        # move output movie particle stacks and change rlnImageName in star file
+        from convert import relionToLocation, locationToRelion
+        for objId in mdAll:
+            index, imgPath = relionToLocation(mdAll.getValue(md.RLN_IMAGE_NAME, objId))
+            baseFn = os.path.basename(imgPath)
+            newPath = pwutils.join(self._getExtraPath('output'), baseFn)
+            newLoc = locationToRelion(index, newPath)
+            if not pwutils.exists(newPath):
```
delarosatrevin: There are many functions here (that sum up for performance penalty) that only make sense for particles stacks, not for each movie-particle. You could compute a dict in the previous loop and use the mapping here.

pyworkflow/em/packages/relion/protocol_extract_particles_movies.py
```diff
@@ -0,0 +1,431 @@
...
+            particleId, _ = relionToLocation(mdAll.getValue(md.RLN_PARTICLE_ORI_NAME, objId))
+            mdAll.setValue(md.MDL_PARTICLE_ID, long(particleId), objId)
+
+        # delete old imgPath
+        pwutils.cleanPath(self._getExtraPath('output/extra'))
+
+        particleMd = self._getFileName('outputParts')
+
+        mdAll.write(particleMd)
+        readSetOfMovieParticles(particleMd, movieParticles,
```
delarosatrevin: Another optimization could be to avoid writing the star file and later parsing it in readSetOfParticles. We could try to directly generate the SetOfMovieParticles directly. I guess this could reduce some time.

pyworkflow/em/packages/relion/convert.py
azazellochg: yes, these are due to the fact that priors are removed in 1.2 but not in devel

azazellochg:
hi @pconesa I have updated the link, should work now

pconesa:
it does!, thanks. I'm putting it in the server now.

pconesa:
dataset is up

pyworkflow/em/packages/relion/protocol_extract_particles_movies.py
azazellochg: true, the problem is the following: relion reads input movie and ptcls star files. rlnMicrographMovieName from movie star should equal rlnImageName from ptcls star  + "_movie.mrcs" otherwise it cannot match the two files. The output star from relion will be rlnMicrographMovieName + "_extract.star". Maybe you can come up with a general solution to sort this out?

azazellochg: _aligned_movie.mrcs comes from movie alignment protocols (all of them). But of course we should consider the case when user just imported aligned movies.

azazellochg: In xmipp movie extr protocol all input movies are renamed to movie_000001 etc but I do not like it because we the original name is lost, the only thing linking output movie particles to particles is particleId, but it's not very helpful, I would prefer micname or smth..

azazellochg:
Closing this PR, since @delarosatrevin has his own branch for this PR.

delarosatrevin:
We can merge my branch back to here. ;)

azazellochg:
@pconesa , yes this should be the case. This PR is based on devel, so there are no release-1.2.1 changes here

azazellochg:
Ok, there are few things left to add:

- [ ] rlnAverageNrOfFrames should be saved in output movie particle set (default=1 if running xmipp protocol)
- [ ] the same number should be passed to star file in protocol_base [here](https://github.com/azazellochg/scipion/blob/az-relion-movie-ptcls/pyworkflow/em/packages/relion/protocol_base.py#L871)
- [ ] check if the whole pipeline works if using xmipp

delarosatrevin:
HI @azazellochg, what is the state of this PR? Do you need something from my side?

azazellochg:
Hi @delarosatrevin sorry it takes me so long to review this. I think it works and can be merged into devel now. If anything wrong comes up I will fix it in devel-fixes branch

delarosatrevin:
@azazellochg not need to sorry at all!!! I just wanted to check there was nothing pending on my side. Ok, let's merge it then and if any issue show up, address it at that moment.

PR #1545, opened 2018-03-27
Relion 3D classification with multiple references

ajimoreno:
The Relion 3D classification protocol has been modified to be able to manage a set of volumes as reference 3D maps. For it, few changes have been made in protocol_classify3d and protocol_base from Relion.

### DIFF
--- a/pyworkflow/em/packages/relion/protocol_base.py
+++ b/pyworkflow/em/packages/relion/protocol_base.py
@@ -194,12 +194,24 @@ class ProtRelionBase(EMProtocol):
                            'the first iteration.')
         group = form.addGroup('Reference 3D map',
                               condition='not doContinue and not is2D')
+        group.addParam('useMultipleVolumes', BooleanParam, default=False,
+                      label="Use multiple volumes")
         group.addParam('referenceVolume', PointerParam, pointerClass='Volume',
                        important=True,
                        label="Input volume",
-                       condition='not doContinue and not is2D',
+                       condition='not doContinue and not is2D and not '
+                                 'useMultipleVolumes',
                        help='Initial reference 3D map, it should have the same '
-                           'dimensions and the same pixel size as your input '
+                            'dimensions and the same pixel size as your input '
+                            'particles.')
+        group.addParam('referenceVolumeSet', PointerParam,
+                       pointerClass='SetOfVolumes',
+                       important=True,
+                       label="Set of multiple input volumes",
+                       condition='not doContinue and not is2D and '
+                                 'useMultipleVolumes',
+                       help='Initial reference 3D maps should have the same '
+                            'dimensions and the same pixel size as your input '
                             'particles.')
         group.addParam('isMapAbsoluteGreyScale', BooleanParam, default=False,
                        label="Is initial 3D map on absolute greyscale?",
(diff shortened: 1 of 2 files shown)


delarosatrevin:
I have addressed the same functionality here: 1c2445410b4c6ed6a8b91356aed65b4cbfe5aab5. 
Some differences:
* Same input is used for single volume, multiple volumes changing the pointerClass param
* Also references for 2D are allowed
* References are converted to MRC if not so, Relion does not support all formats, so creating the .star file is not enough
* This option is only available when classification
* Moreover, if the references have not the same dimensions, they are scaled to match the particles...so an extra resize run is avoided

pyworkflow/em/packages/relion/protocol_base.py
```diff
@@ -194,12 +194,24 @@ def _defineParams(self, form):
                            'the first iteration.')
         group = form.addGroup('Reference 3D map',
                               condition='not doContinue and not is2D')
+        group.addParam('useMultipleVolumes', BooleanParam, default=False,
+                      label="Use multiple volumes")
         group.addParam('referenceVolume', PointerParam, pointerClass='Volume',
                        important=True,
                        label="Input volume",
-                       condition='not doContinue and not is2D',
+                       condition='not doContinue and not is2D and not '
```
delarosatrevin: I think that multiple references are only allowed for classification, not refinement, but not sure.

pyworkflow/em/packages/relion/protocol_base.py
```diff
@@ -973,16 +985,29 @@ def _setNormalArgs(self, args):
...
         if self.IS_CLASSIFY:
             args['--K'] = self.numberOfClasses.get()
             if self.limitResolEStep > 0:
                 args['--strict_highres_exp'] = self.limitResolEStep.get()
     
         if self.IS_3D:
             if not self.IS_3D_INIT:
-                args['--ref'] = convertBinaryVol(self.referenceVolume.get(),
-                                                 self._getTmpPath())
+                if self.useMultipleVolumes:
```
delarosatrevin: Conversion should not be done here! The conversion before of the single volume was out of place...this function is only for defining the parameters...but any work (neither conversions) should be done inside this function.

pyworkflow/em/packages/relion/protocol_base.py
```diff
@@ -973,16 +985,29 @@ def _setNormalArgs(self, args):
...
+                    starFile = open(self._getExtraPath('references.star'),'w')
+                    starFile.write('data_references\n')
```
delarosatrevin: We have classes and functions to operate with Metadata!!!! (including Xmipp xmd, Relion .star and Bsoft .star). So, why to hand-write the .star file here?

ajimoreno:
@delarosatrevin I'm sure that your code is more complete and has more functionalities, so I think it will be better to use it. If you find something useful in this branch, just take it. 
Anyway, we will soon need to use the relion 3d classification with multiple references, so please upload this functionality when you can.

delarosatrevin:
Hi @ajimoreno, I just wanted to point out that I have implemented the same functionality...and also commented on some problems in the implementation here. My changes are somehow isolated in the commit previously referred, but they are in a branch with other changes. I will not have time to merge my changes to devel and test that everything is working there.

cossorzano:
Maybe the best is that we then start from our commit, go over the suggestions you are doing, and when you merge your branch we leave a single piece of code with the merge of the two.

delarosatrevin:
That's fine with me.

josuegbl:
I create a PR ( #1547) with @delarosatrevin's changes. I think this PR can be removed.

pconesa:
I'm cancelling this since it is already inreñease-1.2.1

PR #1564, opened 2018-04-12
hida tab in form if all params are expert and normal mode is active.

### LINKED ISSUE #1558: hide Section Tab(form) if all options are advanced
Now, it is impossible hide a section even if  expertLevel is set to LEVEL_ADVANCED.

pconesa:
Closes #1558

### DIFF
--- a/pyworkflow/gui/form.py
+++ b/pyworkflow/gui/form.py
@@ -1574,25 +1574,26 @@ class FormWindow(Window):
         r = 0
         sectionsFrame = tk.Frame(parent) 
         configureWeigths(sectionsFrame)
-        tab = ttk.Notebook(sectionsFrame) 
-        tab.grid(row=0, column=0, sticky='news',
+        tabs = ttk.Notebook(sectionsFrame)
+        tabs.grid(row=0, column=0, sticky='news',
                  padx=5, pady=5)
         self._sections = []
         
         for section in self.protocol.iterDefinitionSections():
             label = section.getLabel()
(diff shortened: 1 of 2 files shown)


pyworkflow/gui/form.py
```diff
@@ -1899,9 +1901,19 @@ def _checkChanges(self, paramName):
     def _checkAllChanges(self):
         for paramName in self.widgetDict:
             self._checkCondition(paramName)
-            
+
+    def _checkTabsVisibility(self):
+        for index, sectionW in enumerate(self._sections):
+            tabs = sectionW.master
+            if sectionW.section.isExpert():
```
delarosatrevin: @pconesa I don't want to add more work on you...but would be good to make this more consistent with the visibility of the parameters. What I mean, is that you should check not only for the Expert value, but also for the condition (same as for params) to decide when to show/hide a tab. 

Nice work anyway!

pyworkflow/gui/form.py
```diff
@@ -1899,9 +1901,19 @@ def _checkChanges(self, paramName):
...
+                if self.protocol.expertLevel.get() == params.LEVEL_NORMAL:
```
delarosatrevin: .get() is not needed for comparision

delarosatrevin:
Yes (I guess this is checked anyway, we should avoid double evaluation) and also if there is a condition for the Tab itself.

pconesa:
I'm closing this until I've time to implement missing things.

PR #1566, opened 2018-04-17
incorporate workflow repository to python

rmarabini:
This pull request incorporates a few lines of code in Scipion so it can connect to the Scipion Workflow Repository.

The workflow repository is available at URL: http://workflows.scipion.i2pc.es/
There is a help page at: https://github.com/I2PC/scipionWorkflowRepository/wiki/Scipion-Workflow-Repository-Help-Page.

No test has been created since I do not know how to test the tk part of Scipion.

You may start the search GUI from Scipion (select project from the menu and then search workflow) or connect directly to the search URL at http://workflows.scipion.i2pc.es. 

To upload a protocol using Scipion select the protocols (click on them while holding the CTRL-key) and then right click and choose upload workflow from the pop up menu. At this point a browser will be opened, fill the form (all fields are compulsory) and press the create button. Thanks for share your workflows ;-).

---------------

Totally unrelated with the above describe problem I also fix a small bug in volume import protocol when a ·D map is imported but not copied and the 3Dmap format is not mrc/ccp4

### DIFF
--- a/pyworkflow/em/protocol/protocol_import/volumes.py
+++ b/pyworkflow/em/protocol/protocol_import/volumes.py
@@ -127,7 +127,6 @@ class ProtImportVolumes(ProtImportImages):
         volSet.setSamplingRate(samplingRate)
 
         for fileName, fileId in self.iterFiles():
-            newFileName = abspath(self._getVolumeFileName(fileName))
             x, y, z, n = imgh.getDimensions(fileName)
             if fileName.endswith('.mrc') or fileName.endswith('.map'):
                 fileName += ':mrc'
@@ -136,6 +135,8 @@ class ProtImportVolumes(ProtImportImages):
                     n = 1
                 else:
                     zDim = z
+            else:
+                zDim = z
             origin = Transform()
             if setOrigCoord:
                 origin.setShiftsTuple(self._getOrigCoord())
@@ -147,10 +148,13 @@ class ProtImportVolumes(ProtImportImages):
             vol.setOrigin(origin)  # read origin from form
 
             if self.copyFiles or setOrigCoord:
+                newFileName = abspath(self._getVolumeFileName(fileName, "mrc"))
                 adaptFileToCCP4(fileName, newFileName, origin.getShifts(),
                                 samplingRate,
                                 ORIGIN)
             else:
+                newFileName = abspath(self._getVolumeFileName(fileName))
+
                 if fileName.endswith(':mrc'):
                     fileName = fileName[:-4]
                 createAbsLink(fileName, newFileName)
@@ -197,8 +201,12 @@ class ProtImportVolumes(ProtImportImages):
                            (self._getVolMessage(), self.samplingRate.get()),)
         return methods
 
-    def _getVolumeFileName(self, fileName):
-        baseFileName="import_" + basename(fileName).split(".")[0] + ".mrc"
+    def _getVolumeFileName(self, fileName, extension=None):
+        if extension is not None: 
+            baseFileName="import_" + basename(fileName).split(".")[0] + ".%s"%extension
+        else:
+            baseFileName="import_" + basename(fileName)
+
         return self._getExtraPath(baseFileName)
 
     def _getOrigCoord(self):
--- a/pyworkflow/em/protocol/protocol_import/volumes.py
+++ b/pyworkflow/em/protocol/protocol_import/volumes.py
@@ -205,7 +205,7 @@ class ProtImportVolumes(ProtImportImages):
         if extension is not None: 
             baseFileName="import_" + basename(fileName).split(".")[0] + ".%s"%extension
         else:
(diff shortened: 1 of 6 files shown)


pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1826,6 +1833,62 @@ def _export(obj):
...
+            # connection
+            datagen, headers = poster.encode.multipart_encode(params)
+            opener = poster.streaminghttp.register_openers()
+            # create request and connect to server
+            response = opener.open(urllib2.Request(upload_file_url,
+                                   datagen, headers))
+            # server returns dictionary as json with remote name of the saved file
+            _dict = json.loads(response.read())
+            # server has stored file in file named _dict['jsonFileName']
+            # TODO: This version thing is a horrible hack but it is not
```
pconesa: Really?

rmarabini: I do not know how to answer to the above comment. Should I do something?

Ideally scipion should offer an easy way to get the version something like:

from scipion import version
version = version()

With the current code the only way I know how to get the version is:

+            import imp
 +            currentDir = os.path.dirname(os.path.abspath(__file__))
 +            currentFile = os.path.join(currentDir,'../../../scipion')
 +            mod = imp.load_source('scipion', currentFile)
 +            version = mod.getVersion(False)

the above code is rather ugly

pyworkflow/em/protocol/protocol_import/volumes.py
```diff
@@ -197,8 +201,12 @@ def _methods(self):
                            (self._getVolMessage(), self.samplingRate.get()),)
         return methods
 
-    def _getVolumeFileName(self, fileName):
-        baseFileName="import_" + basename(fileName).split(".")[0] + ".mrc"
+    def _getVolumeFileName(self, fileName, extension=None):
+        if extension is not None: 
+            baseFileName="import_" + basename(fileName).split(".")[0] + ".%s"%extension
```
delarosatrevin: Need some spaces for PEP8

pyworkflow/gui/project/constants.py
```diff
@@ -41,3 +41,8 @@
                STATUS_SCHEDULED: '#F3F5CB'
                }
 
+WORKFLOW_REPOSITORY_SERVER='http://workflows.scipion.i2pc.es/'
```
delarosatrevin: I don't really like this values in the python code...it should be in some .conf file (and for backward compatibility with some reasonable default)

pyworkflow/gui/project/project.py
```diff
@@ -39,7 +39,7 @@
 import uuid
 import SocketServer
 import tempfile
-
+import webbrowser
```
delarosatrevin: Can this new dependency be moved down to the function where it is used? Or it is pretty much standard and it is not a problem?

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -55,6 +56,9 @@
 
 from constants import STATUS_COLORS
 from pyworkflow.gui.project.utils import getStatusColorFromNode
+import poster.streaminghttp
```
delarosatrevin: I don't like spreading more and more dependencies for implementing new functionality. Can this be wrapped into a function? (in another module)

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1826,6 +1833,62 @@ def _export(obj):
             entryLabel='File', entryValue='workflow.json')
         browser.show()
 
+    def _exportUploadProtocols(self):
```
delarosatrevin: Please make sure that no error will happen (or at least that will not interfere with the current processing) in the GUI if case there is not internet connnection or there is any failure

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1826,6 +1833,62 @@ def _export(obj):
...
+
+        def uploadWorkflow(upload_file_url, upload_metadata_url, jsonFileName):
```
delarosatrevin: Maybe this whole function (and its dependencies and imports) can be moved to a separate module / place

delarosatrevin:
@rmarabini I think that maybe we could introduce this  feature for release-1.2.1, what do you think? In my opinion, if we address the dependencies that I pointed out in the comments, it is not much a risk and we could release it earlier.

rmarabini:
OK, let us sit together tomorrow and think what need to be modified

            Roberto

pyworkflow/gui/project/constants.py
delarosatrevin: Or maybe (discussing with @rmarabini ) a config.py file

rmarabini:
I close this pull request since it is going to be backported to Scipion 1.2

PR #1576, opened 2018-04-27
Need to split on os.pathsep (":") when doing appendUnique of LIBPATH to libPathCopy.

akesandgren:
Need to split on os.pathsep (":") when doing appendUnique of LIBPATH to libPathCopy.

Otherwise the end result will be
-Lpath1:path2:path3
instead of
-Lpath1 -Lpath2 -Lpath3.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -464,7 +464,7 @@ def addProgram(env, name, src=None, pattern=None, installDir=None,
     cxxflagsCopy = cxxflags + env['CXXFLAGS']
     linkflagsCopy = linkflags + env['LINKFLAGS']
     ldLibraryPathCopy = [env['LIBPATH']]
-    appendUnique(libPathsCopy, env.get('LIBPATH', ''))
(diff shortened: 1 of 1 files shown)


delarosatrevin:
Thanks @akesandgren! Do you think it would be good to review all usages of appendUnique function?

akesandgren:
Maybe but this is the only one i had problems with so far.
But any use of environment variable using ":" separator -> python list needs a split i guess.

delarosatrevin:
@pconesa I would propose to incorporate this fix for the next 1.2.1 release...what do you think? (Initially only in that specific place in the code)

pconesa:
Ok, @akesandgren , do you mind to make the PR to release-1.2.1 branch instead. or I can take that line there.

delarosatrevin:
@pconesa I can cherry pick this change to that branch.

pconesa:
@akesandgren , this is already incorporated in another branch. Thank you both!

PR #1582, opened 2018-05-06
backporting scipion access to repository web site

rmarabini:
Code needed by scipion to access to the workflow repository.

Help is here: https://github.com/I2PC/scipionWorkflowRepository/wiki/Scipion-Workflow-Repository-Help-Page

URL to repository is: http://workflows.scipion.i2pc.es/

NOTE: fill free to play with the repository, you may delete any workflow uploaded by you and you may NOT delete workflows uploaded by others ;-)


pyworkflow/gui/project/notifier.py
```diff
@@ -33,8 +33,10 @@
 from datetime import timedelta, datetime
 
 import pyworkflow.utils as pwutils
-SCIPION_STATS_SERVER = 'http://scipion.i2pc.es'
-SCIPION_STATS_WORKFLOW_APP = SCIPION_STATS_SERVER + '/report_protocols/api/workflow/workflow/'
+
+import sys
+sys.path.append(os.path.join(os.environ['SCIPION_HOME'], 'config'))
```
delarosatrevin: Why this manipulation of the sys.path? I think it is not necessary since the from config.config should work. Or maybe we need to add a proper __init__.py module to config that explicitly export the symbols that should be visible from outside this module.

pyworkflow/gui/project/project.py
```diff
@@ -39,6 +39,7 @@
 import uuid
 import SocketServer
 import tempfile
+import webbrowser
```
delarosatrevin: Is this import still necessary here? It is not used (I assume is inside the imported function or module)

pyworkflow/gui/project/project.py
```diff
@@ -55,6 +56,10 @@
 
 # Import possible Object commands to be handled
 from base import ProjectBaseWindow, VIEW_PROTOCOLS, VIEW_PROJECTS
+from workflowRepository import searchWorkflow
```
delarosatrevin: This module name does not follows our standard: filenames should be lowercase and separated by underscore.

pyworkflow/gui/project/project.py
```diff
@@ -55,6 +56,10 @@
...
+
+sys.path.append(os.path.join(os.environ['SCIPION_HOME'], 'config'))
```
delarosatrevin: I think this is not a good approach to manipulate the sys.path

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -55,7 +54,7 @@
 
 from constants import STATUS_COLORS
 from pyworkflow.gui.project.utils import getStatusColorFromNode
-
+from workflowRepository import exportUploadProtocols
```
delarosatrevin: Again here the workflowRepository name does not follow our conventions.

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -1802,6 +1804,13 @@ def _export(obj):
             entryLabel='File', entryValue='workflow.json')
         browser.show()
 
+    def _exportUploadProtocols(self):
+        protocols = self._getSelectedProtocols()
+        project = self.project
```
delarosatrevin: Is it validated if an Error occurs? (e.g. that there is not internet connection or something else)

pyworkflow/gui/project/workflowRepository.py
```diff
@@ -0,0 +1,77 @@
+import os, sys
+import poster.streaminghttp
+import urllib2
+import webbrowser
+import json
+import tempfile
+
+sys.path.append(os.path.join(os.environ['SCIPION_HOME'], 'config'))
```
delarosatrevin: Avoid this.

pyworkflow/gui/project/project.py
rmarabini: +import webbrowser has been removed

pyworkflow/gui/project/project.py
rmarabini: module has been renamed

pyworkflow/gui/project/project.py
rmarabini: The only possibilities I may think are:

1) add a __init__.py file at the top level
2) put the config.py file in the pyworkflow directory (or any of the subdirectories)

I am happy with any of the solutions. In case you like option (2) please specify the subdirectory.

pyworkflow/gui/project/viewprotocols.py
rmarabini: fixed

pyworkflow/gui/project/viewprotocols.py
rmarabini: Let see.

   If there is any problem connecting the system will show a pop up window with the Error message:   "Error connecting to workflow repository" followed by the error message given by the system

 (1)  If there is no internet connection in my machine a get :  "open error: failure in name resolution"

(2) If the web site is down  inmy machine a get "HTTP Error 404: Not Found"

I cannot think in other cases

pyworkflow/gui/project/workflowRepository.py
rmarabini: this problem has been already addressed above

rmarabini:
So far I do not have more comments.  In summary either we place a __init__,py  in $SCIPION_HOME or we change the config.py to another place under the pyworkflow directory.

your turn @delarosatrevin ;-)

pyworkflow/gui/project/notifier.py
rmarabini: No, it will not work unless we create a __init__.py file in $SCIPION_HOME.
So either we create the __init__.py file or we move config.py under the pyworkflow directory three. If you like the second option better suggest a place to put config.py

PR #1588, opened 2018-05-07
Reduce pseudoatoms

JaviMota:
Short Xmipp program to reduce the number of pseudoatoms according to their intensities. 
New Xmipp file created pdb_reduce_pseudoatoms.


cossorzano:
We will add an option to select by number of atoms too

Issue #1571, opened 2018-04-23
em.data.test_data.TestImage - testOrigin FAILS


dmaluenda:
`scipion test tests.em.data.test_data.TestImage` FAILS in devel branch since the mic under tests has not sampling rate.

The problem is in the function [_getDefaultOrigin()](https://github.com/I2PC/scipion/blob/daa413e7887bd6dad8481ee5db4d7c0311f1aa79/pyworkflow/em/data.py#L596), maybe it needs to assert if the Image have sampling rate and if not then do something.

Any idea? @mmmtnez

dmaluenda:
Ok! So, the best option is to introduce a sampling rate to the micrograph by mic.setSamplingRate(1) in the test??

josuegbl:
I dont know how works the test, but if it creates a new micrograph object, the sampling rate must be set. Check with @mmmtnez and @rmarabini all the details about this, please.

PR #1589, opened 2018-05-07
Bug fixing related with buildbot (and data test)

dmaluenda:
Some bug are fixing:
 - We avoid passing the PYTHONPATH env variable of the user to scipion enviroment to avoid that pip looks for packages out of scipion.
 - CUDA enviroment variables now are pointing to `/usr/local/cuda/lib64` to make it version-independent. In addition, some bug fixing in `scipion` script.
 - The pip instalation is done with `--no-setuptools` to after install a certain version of setuptools to do it more robust.
 - we assert that a volume have sampling rate in `em/data.py` 
 - bug fixing in `tests/em/data/test_data.py`


pyworkflow/em/data.py
```diff
@@ -595,6 +595,8 @@ def getOrigin(self, force=False):
 
     def _getDefaultOrigin(self):
         sampling = self.getSamplingRate()
+        if self.getSamplingRate() is None:
```
josuegbl: I strongly disagree with this. The sampling rate is one of the most important parameters that we store for almost all EM Objects. This kind of "fixes" could hide important bugs in the code. If you dont have sampling Rate for Volumes in certain protocol(s), YOU MUST FIX THE PROTOCOL(s). Please, to modify this type of core files in Scipion we need to have consensus and think very well the implications of the changes.

delarosatrevin: I totally agree with @josuegbl on this point.

dmaluenda:
We reviewed with @rmarabini as mentioned [here](https://github.com/I2PC/scipion/issues/1571), but maybe the best option is setting the sampling rate by hand in the test (now is repushed with this option).

Thanks.

scipion
```diff
@@ -203,19 +203,21 @@ try:
         [join(SCIPION_SOFTWARE, 'bin'),
          BUILD['JAVA_BINDIR'],
          BUILD['MPI_BINDIR'],
+         BUILD['CUDA_BIN'],
```
pconesa: @cossorzano , this also might be needed in your branch.

scipion
```diff
@@ -203,19 +203,21 @@ try:
...
          os.environ.get('PATH', '')]
     )
     LD_LIBRARY_PATH = os.pathsep.join(
         [join(SCIPION_SOFTWARE, 'lib'),
          BUILD['MPI_LIBDIR'],
-         BUILD['CUDA'],
+         BUILD['CUDA_LIB'],
```
pconesa: and this too

pyworkflow/em/data.py
pconesa: thanks, @josuegbl and @delarosatrevin for keeping and eye on this.

pconesa:
I guess all is fixed now..

pyworkflow/em/data.py
rmarabini: I guess it make sense to modify: "pyworkflow/tests/em/data/test_data.py" and add in test function"testOrigin"  the line "mic.setSamplingRate(1.)" so the function looks like:


    def testOrigin(self):
        fn = self.mic1
        mic = Micrograph()
        mic.setFileName(fn)
       mic.setSamplingRate(1.) ##################new line
        t = mic.getOrigin(True)
        referenceT = Transform()
        referenceT.setShifts(4608.,4720.,1)
        mt = t.getMatrix()
        mreferenceT = referenceT.getMatrix()
        for i in range(4):
            for j in range(4):
                self.assertAlmostEquals(mt[i][j],mreferenceT[i][j],1)

@dmaluenda

PR #1595, opened 2018-05-08
Adding LocScale (new protocol from new package)

dmaluenda:
A new protocol/package from [Arjen J Jakobi](https://elifesciences.org/articles/27131) is added. It runs under Eman eviroment (i.e. import Eman modules).

It basically "computes contrast-enhanced cryo-EM maps by local amplitude scaling using a reference model".

Inputs: EM-volume, Model-volume, patchSize and (optionally) 3D-Mask
Output: sharpened_volume


josuegbl:
I propose change the name of the package by SachseLab (in accordance as we have).  I also propose you to implement a convertInputStep cause, maybe, you should convert certain volumes cause its formats are not supported by the program.

dmaluenda:
I included the suggestions.

Thanks Josue!

dmaluenda:
I included the suggestions.

Thanks Josue!

PR #1599, opened 2018-05-11
handler for PDB/MMCIF files

rmarabini:
This code is able to read/write atomic structure files from/to disk and convert among PDB/MMCIF (The aim is similar to imageHandler but for atomic structure files)

dependencies: biopython module
Tests: ./scipion test tests.em.data.test_convert_atom_struct.TestAtomicStructHandler


rmarabini:
@cossorzano  you may want to review this pull request

PR #1600, opened 2018-05-12
 backporting scipion access to repository web site SECOND version

rmarabini:
After an in-deep refactoring made mainly by @delarosatrevin,  this is a new pull request to  backport scipion access to the workflow repository web site

**Help** is here: https://github.com/I2PC/scipionWorkflowRepository/wiki/Scipion-Workflow-Repository-Help-Page

URL to **repository** is: http://workflows.scipion.i2pc.es/

NOTE: fill free to play with the repository, you can delete any workflow uploaded by you and you canNOT delete workflows uploaded by others ;-)

**QUESTION**: It would be nice to have a version and a date tags  in the workflow json files produced by scipion. I do not mind to introduce it  but I am afraid that it will brake some code, in particular the java script that draws the workflows. Should I go ahead and introduce something like:

 {
  "created_at": "Thu Jun 22 21:00:00 +0000 2017",
  "scipion.version": 1.2
 }


delarosatrevin:
@rmarabini as you already said, modifying the .json could (I'm not completely sure about this) break importing .json. So, I would propose to evaluate (and maybe implement) this change for a major release, not for the upcoming 1.2.1.

Since I have made some changes over your original PR, I think it would be better that someone else take a look now. (e.g, @pconesa ;)

PR #1602, opened 2018-05-16
Release 1.2.1 dm4

pconesa:


pyworkflow/em/convert.py
```diff
@@ -157,22 +168,15 @@ def convert(self, inputObj, outputObj, dataType=None, transform=None):
         inputLoc = self._convertToLocation(inputObj)
         outputLoc = self._convertToLocation(outputObj)
         
-        if (inputLoc[1].lower().endswith('dm4') or
```
delarosatrevin: @pconesa We still need the eman2 hack for .img (imagic) format.

software/em/xmipp/java/src/xmipp/.classpath
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
delarosatrevin: This file should not be included...I guess it is an IDE generated file.

pyworkflow/em/convert.py
```diff
@@ -148,6 +148,17 @@ def existsLocation(cls, locationObj):
         
         return os.path.exists(fn)
     
+    @classmethod
+    def adaptDataType(cls, inDataType, outputFilename):
```
delarosatrevin: Comments (in the function docs) about what this function does? Maybe a better name for the function?

Apart from that, I'm not sure if the conversion to DT_FLOAT (in case of DT_SCHAR) is justified. MRC format supports signed and unsigned integers, which is a more close datatype to DT_SCHAR than float.

software/em/xmipp/java/src/xmipp/.classpath
pconesa: Actually I removed it.

delarosatrevin: Ok, I'm sorry, I was confused and thought it was added.

pyworkflow/em/convert.py
pconesa: Thank Jose....you suggest to change data type to DT_USHORT?

delarosatrevin: I would suggest to change to the closer one supported by the file format (I'm not sure if DT_USHORT in this particular case of MRC)

pconesa: is convertDataType a better name? Suggestions are welcome.

delarosatrevin: This is not accurately neither...you are not doing any conversion.

What about "getSupportedDataType" ?? (and provide documentation about what the function does)

pyworkflow/em/convert.py
pconesa: Yep, that is true.

pconesa:
Why is this? I didn't  merge the PR!!!

pconesa:
oh....I just pushed to I2PC instead!! grrr...well all your (@delarosatrevin ) suggestions were addressed.

pyworkflow/em/convert.py
josuegbl: I totally agree with @delarosatrevin .

PR #1605, opened 2018-05-18
Reduce number of pseudoatoms

JaviMota:
Short script to reduce the number of pseudoatoms in case of memory error


rmarabini:
Hi @pconesa,

   What problem do you see in  pdb.cpp?

pconesa:
he,he the main problem is me!!! I can't tell if those changes are affecting existing functionality. If you don't seen any problem....I think we can o ahead.

Issue #1610, opened 2018-05-20
Memory leak in protocols using pyplot.Figure


matthutchison:
`pyplot.Figure` references are retained after creating them and are not gc'd unless the `Figure` explicitly closed, leading to a slow accumulation of reserved memory in protocols like motioncorr2.  It's especially noticeable in streaming where the protocol instance doesn't end until the streaming session ends. This appears to be intentional on the part of matplotlib (see note here in the [pyplot.figure docs](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.figure.html)).

PID 22422 in the images below is a motioncorr2 protocol in streaming mode that's been running for ~33 hours and accumulated ~5.5GB of reserved memory.

<img width="1345" alt="screen shot 2018-03-21 at 9 02 10 am" src="https://user-images.githubusercontent.com/4432140/40283657-b27b8318-5c4f-11e8-9628-af90421431a8.png">
<img width="543" alt="screen shot 2018-03-21 at 9 04 50 am" src="https://user-images.githubusercontent.com/4432140/40283658-baed6778-5c4f-11e8-8459-52164331b56f.png">

I haven't explored the full implications of this throughout, but there are some instances that look safe to fix by just closing the `Figure`.  I'll add a PR shortly for any instance where (like motioncorr2) the `Figure` is local to the plotting function, `savefig` is called, and the `Figure` instance is not returned.

PR #1611, opened 2018-05-20
Some fixes for https://github.com/I2PC/scipion/issues/1610

matthutchison:
In any instance where
- The `Figure` instance is created locally
- `savefig` is called
and
- The `Figure` is not returned from the function

It should be safe to immediately call .close() on the Figure instance.

There are still `savefig` calls that may be easy corrections in:
- pyworkflow/em/packages/xmipp3/protocol_multireference_alignability.py - I think the reference is supposed to be retained here?
- pyworkflow/em/packages/resmap/protocol_resmap.py - I'm not sure about this one
- pyworkflow/web/app/views_util.py - plotter passed in externally (fixed one instance I think safely, but pyworkflow/web/app/em_viewer.py is still using with a view call that I’m not familiar with the implications of)

This implemented as a test fix for #1610 results in no leak (image below shows ~1600 movies processed using the same set of protocols as in the original issue, but now without any of the accumulation that we saw before).

screen shot 2018-05-20 at 4 41 02 pm

### DIFF
--- a/pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
+++ b/pyworkflow/em/packages/grigoriefflab/protocol_unblur.py
@@ -466,6 +466,7 @@ eof
         first, _ = self._getFrameRange(movie.getNumberOfFrames(), 'align')
         plotter = createGlobalAlignmentPlot(shiftsX, shiftsY, first)
         plotter.savefig(self._getPlotGlobal(movie))
+        plotter.close()
 
 def createGlobalAlignmentPlot(meanX, meanY, first):
     """ Create a plotter with the shift per frame. """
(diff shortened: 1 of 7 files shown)


delarosatrevin:
Dear @matthutchison, thanks a lot for this fix! I also observed a memory leak in motioncor2 (scipion protocol) but I didn't had the time to dig deeper the problem. Also the optical-flow protocol was having some issues related to the Figure in matplotlib. This bug that you has found (and fixed) is very very important. 

Would you mind to do the PR against branch release-1.2.1? We are planning to release a new minor version for the end of June, so your fix will be available there.

delarosatrevin:
I'm closing this one since I ported this PR to release-1.2.1 branch. Thanks again @matthutchison!

PR #1618, opened 2018-05-23
Release 1.2.1 relion fixes

delarosatrevin:
* Fixed bug when using Relion picking: the run of the protocol for optimize the picking params was generating a set of micrographs (subset based on defocus) and a set of coordinates (linked to the subset), but when executing for the whole set, still the output coordinates was there and with bad reference to microgasph!!! Solution: change the name to 'outputXXXSubset'. Still, it seems to be a deeper problem with deleting outputs of a protocol in a Continue mode.
* Alleviated the problem of selecting input CTF and getting no input in the select dialog...now a refresh param is passed as True and it seems to be less problematic now, not sure if will pick up the inputs immediately. The downside is that it takes a bit more of time when opening, but I think is a good price to pay for avoiding the annoyance of not finding any input CTF.

### DIFF
--- a/pyworkflow/em/packages/relion/protocol_autopick_v2.py
+++ b/pyworkflow/em/packages/relion/protocol_autopick_v2.py
@@ -369,6 +369,20 @@ class ProtRelion2Autopick(ProtParticlePickingAuto, ProtRelionBase):
             writeReferences(self.getInputReferences(),
                             self._getPath('input_references'), useBasename=True)
 
+        # FIXME: (JMRT-20180523) The following code does not seems to work
+        # here it has been worked around by changing the name of the wizard
+        # output but this seems to reflect a deeper problem of deleting
+        # already existing output objects in a protocol. Maybe when updating
+        # from run.db to project.sqlite?
+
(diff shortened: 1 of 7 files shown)


azazellochg:
I won't be able to review this until Tuesday

delarosatrevin:
This PR should not be merged yet. I'm investigating some other issues with Relion picking wizard...not sure if related to my changes, but I will submit the changes here.

delarosatrevin:
Bug fixed! A tricky one. The set used for creating the output (based on defocus subset) change because the whole input was not the same (because of streaming) than when the createOutputStep was executed and created the input_micrographs.star for relion program. 

Just a reminder that "streamifying" protocols requires some effort and a lot of testing. 

PR ready to be reviewed again. ;)

pconesa:
Can we merge this?

delarosatrevin:
@azazellochg are you working on this issue? or you want me to take a look?

azazellochg:
@delarosatrevin It would be great if you could have a look, I am not very familiar with xmipp extract movie particles protocol. It looks very tied to movieId (as in protocol_movies)..

PR #1620, opened 2018-05-24
Tolerate loading errors

### LINKED ISSUE #1597: pop-up raised with msj:'NoneType' object has no attribute 'hasCondition' (v1.2)
this is happening with protocols that has been opened with a devel-related branch and when I try to open in v1.2 branch, the message 'NoneType' object has no attribute 'hasCondition'.

pconesa:
This is replacing the previous PR.
Closes #1597 

Tolerates loading projects with errors (missing sets sqlites) at loading time and also when creating the summary.

Since now it loads, I would say any "lonely" project.sqlite...I've added a new "gray" color to the boxes when the box has "summaryWarnings" (now either folder missing or a set is missing, quite critical, I would say).

AS @delarosatrevin suggested in the previous PR-.....at loading time, only if the loading error is due to a missing set it will load.I other cases...I guess it will not load the project and will behave as before.

### DIFF
--- a/pyworkflow/project.py
+++ b/pyworkflow/project.py
@@ -1029,7 +1029,14 @@ class Project(object):
             self.runs = self.mapper.selectAllBatch(objectFilter=lambda o: isinstance(o, pwprot.Protocol))
 
             for r in self.runs:
-                self._setProtocolMapper(r)
+
+                # Tolerate loading errors. For support.
+                # When only having the sqlite, sometime there are exceptions here
+                # due to the absence of a set.
+                try:
+                    self._setProtocolMapper(r)
+                except Exception as e:
+                    print("Protocol %s couldn't be loaded. %s" % (r, e))
 
                 # Check for run warnings
                 r.checkSummaryWarnings()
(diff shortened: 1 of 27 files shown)


pyworkflow/project.py
```diff
@@ -995,9 +995,22 @@ def _storeProtocol(self, protocol):
...
     def _setProtocolMapper(self, protocol):
         """ Set the project and mapper to the protocol. """
-        protocol.setProject(self)
-        protocol.setMapper(self.mapper)
-        self._setHostConfig(protocol)
+
+        # Tolerate loading errors. For support.
+        # When only having the sqlite, sometime there are exceptions here
+        # due to the absence of a set.
+        from pyworkflow.mapper.sqlite import SqliteFlatMapperException
```
delarosatrevin: I don't like this import here and the explicit use of SqliteFlatMapper...but I don't see a near future were we implement it more general and other mappers are possible. So, I can live with it for the moment.

pconesa: I agree with that, but didn't want to "generalize the exceptions". Glad that you can also live with that!

PR #1626, opened 2018-05-24
symmetry metrices 

rmarabini:
This pull request  is able to compute the symmetry matrices related with all point symmetries (cyclic, tetragonal, ...). Code has been adapted from chimera. Main difference scipion uses 4x4 matrices while chimera uses 4*3

Tests:
    ./scipion test tests.em.data.test_symmetry.TestSymmetry

I have refactored the  test 
  ./scipion test em.packages.xmipp3.tests.test_protocol_extract_unit_cell

so it uses this new module instead of the file Icosaedron.py (I have deleted this file because it is not longer needed)


josuegbl:
Very good job @rmarabini!

PR #1636, opened 2018-05-30
Add a threshold in phantom_simulate_microscope Test

dmaluenda:
The threshold for the assertion in the phantomSimulateMicroscope test has been increased in order to pass in different branches where the approach has been slightly changed.


pconesa:
thanks @delarosatrevin

PR #1639, opened 2018-06-01
Release 1.2.1 schedule select

delarosatrevin:
The 2d-streamer protocol already introduced the possibility to schedule runs after completion of some previous runs. 

This PR allows the users to select these preRequisites in a simple maner: The list of protocol IDs can be provided and the run will be scheduled...and not started until all preRequisites are finished.

Minor code clean-up was done in form.py

### DIFF
--- a/pyworkflow/gui/form.py
+++ b/pyworkflow/gui/form.py
@@ -52,14 +52,15 @@ from widgets import Button, HotButton, IconButton
 from dialog import showInfo, EditObjectDialog, ListDialog, askYesNo, Dialog
 from canvas import Canvas
 from tree import TreeProvider, BoundTree
-#from pyworkflow.em import findViewers
+
 
 THREADS = 'Threads'
 MPI = 'MPI'
 
-#-------------------- Variables wrappers around more complex objects -----------------------------
 
-class BoolVar():
+# ----------------- Variables wrappers around more complex objects ------------
+
+class BoolVar:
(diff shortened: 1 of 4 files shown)


pconesa:
I think this is a very nice feature, @delarosatrevin ....the only thing I don't like much is asking for IDs?...
Don't you think we could use the object browser there?

delarosatrevin:
Totally agree @pconesa ..but I wanted to keep as simple as possible.
This list sets the preRequisites...that is a propery Protocol class inherits from Step...so it is not a normal param. You can continue to pull the threads...and it can complicate a little bit.
On the other hand...sometimes finding input/ouputs for big projects is a bit challenging.

So, I option by either not allow user to specify it...or allow it in the simplest possible way.

I think we can start with this approach and improved in the future.

pconesa:
Ok, let's have the simplest way with "room" for improvement....

delarosatrevin:
Hi @azazellochg 
1) Yes, there are not relation between the protocol and the waiting for protocols.
2) We need to check this case, it should behave as normal streaming (for which case there is not need to use the "Wait for" feature, since there is an input dependency)

pconesa:
Hi, as a quick and necessary feature I think this is ok...but don't you think this might require a second thought...maybe later.

delarosatrevin:
Yes, sure...and a third and a forth.

PR #1645, opened 2018-06-06
Reduce pseudoatoms with test

JaviMota:
New test file to test pdb_reduce_pseudoatoms


pyworkflow/em/packages/xmipp3/tests/programs/test_programs_xmipp.py
```diff
@@ -775,6 +775,19 @@ def test_case1(self):
                 outputs=["file1.txt","file2.txt","file3.txt","file4.txt"])
 
 
+class PdbReducePseudoatoms(XmippProgramTest):
+    _owner = COSS
```
pconesa: @JaviMota not sure if we will use this in the future but it would be great to add yourself. as owner

delarosatrevin: Yes, maybe better if it could be a list instead of a single entry. What about _maintainers instead of _owner?

delarosatrevin: Maybe we can apply similar rules as in other cases...if the maintainers are defined at package level, use then unless redefined in the protocol.

pconesa: I agree, but we need to refactor this for plugins....we'll do that but later.

PR #1646, opened 2018-06-07
convertThumbnail: has optional flip flags (x, and y)

### LINKED ISSUE #1612: Inverted fitted CTF image in HTML report
I noticed this error while running the course in Umea, but it can be also reproduced with an existing streaming test:
scipion test tests.em.workflows.test_workflow_streaming.TestStreamingWorkflow

The images in the HTML are inverted in Y, different from the one shown in the Showj interface of analyzing results.

image

image

pconesa:
Use flip flags in the HTML report for the mic and PSD
Fix bug in HTML report generation where no ctf protcol is added and alignment hasn't computed the PSD.

Closes #1612

### DIFF
--- a/pyworkflow/em/convert.py
+++ b/pyworkflow/em/convert.py
@@ -398,14 +398,20 @@ class ImageHandler(object):
         """
         return xmipp.FileName(imgFn).isImage()
 
-    def computeThumbnail(self, inputFn, outputFn, scaleFactor=6):
+    def computeThumbnail(self, inputFn, outputFn, scaleFactor=6, flipOnY=False,
+                         flipOnX=False):
         """ Compute a thumbnail of inputFn, save to ouptutFn.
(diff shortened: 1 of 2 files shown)


pconesa:
It also uses
` getConfigPath('execution.summary.html')`
if exists.

delarosatrevin:
@pconesa Despite these changes, still there is a mismatching in the CTF Image (computed with CTFFIND4) as shown below between showj and the web report

image

Anyway, I don't think it is a matter of flipping the image, because I think both viewers (desktop and web) are reading the same image, but maybe displaying it in different ways.

pconesa:
Well we did some effort there...if images to thumbnail are less than 10 we do it straight away otherwise we do it in a thread...but while reviewing this I was wondering if this could have generated steps here...but not sure if a monitor can do that.

PR #1650, opened 2018-06-12
Xmipp CTF estimation with VPP issues fixed and self estimation defocus

JaviMota:
Problems with phase shift label fixed. Propagation of the phase shift label to phase flip and Wiener filter.
Now xmipp ctf estimation is able to estimate the defocus by itself.
CtfEstimateFromPsdFast test done.

### DIFF
--- a/pyworkflow/em/data.py
+++ b/pyworkflow/em/data.py
@@ -122,7 +122,8 @@ class CTFModel(EMObject):
         self._defocusV = Float(kwargs.get('defocusV', None))
         self._defocusAngle = Float(kwargs.get('defocusAngle', None))
         self._defocusRatio = Float()
-        self._phaseShift = Float(kwargs.get('phaseShift', None))
+        self._phaseShift = None if not 'phaseShift' in kwargs \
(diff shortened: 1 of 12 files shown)


pyworkflow/em/data.py
```diff
@@ -123,7 +123,7 @@ def __init__(self, **kwargs):
         self._defocusAngle = Float(kwargs.get('defocusAngle', None))
         self._defocusRatio = Float()
         self._phaseShift = None if not 'phaseShift' in kwargs \
```
delarosatrevin: It is not introduced in this change, but \ for separating lines are not recommended in PEP8 style guide.

It would try to write a shorted version of the line (and more positive):

self._phaseShift = Float(kwargs['phaseShift']) if 'phaseShift' in kwargs else None

I haven't tested the above code or know if fits in 80 wide line

pconesa:
We can't merge any PR that will affect xmipp code (c++ nor python)....until Xmipp split is done. I'm closing this.

PR #1652, opened 2018-06-13
Html report minor fixes

pconesa:
limit x axis of resolution chart to 20
use https for highcharts.js.


pconesa:
wrong base, sorry!

Issue #1631, opened 2018-05-28
xmipp ctf selection finishes prematurely


pconesa:
When input is still in streaming it closes and finishes while input is still running.
![image](https://user-images.githubusercontent.com/785633/40623280-560c4eac-62a6-11e8-979b-836530d7783f.png)

pconesa:
@majtner it this something you streamified?

majtner:
No, I was not working on this one. As far as I know, @ajimoreno was working on streaming process of this protocol.

pconesa:
Ok, sorry! I assign this to her.

ajimoreno:
I'm checking it, but I can't find the problem with the selection protocol. It's running correctly except when it goes after the ctf consensus. I don't know exactly the aim of the ctf consensus, but it seems it gets stuck and the output is not updated and that's the reason why ctf selection stops.

![screenshot from 2018-05-28 20-45-42](https://user-images.githubusercontent.com/26138342/40626366-609f300c-62b8-11e8-90ae-0533b5df6c1f.png)

majtner:
It is possible that there is some problem with ctf consensus. Today, I was looking at consensus protocol and I found some scenarios, when the output was not correct. I will double-check and correct the consensus protocol, maybe it will solve this problem.

pconesa:
Ok, thank you both!

PR #1654, opened 2018-06-14
Merging ctf_discrep. and ctf_selection

majtner:
This PR is merging two protocols (ctf_discrepancy and ctf_selection) to one ctf_consensus. Functionality of both protocols should be preserved. There was introduced new threshold for discarding ctfs based on consensus resolution, which was also added to metadata.
Everything should work also in streaming.
This PR is also solving #1631 
@ajimoreno Please, take a look if all functionality of your ctf_selection is there


pyworkflow/em/data.py
```diff
@@ -128,6 +128,7 @@ def __init__(self, **kwargs):
         self._psdFile = String()
         self._micObj = None
         self._resolution = Float()
+        self._consensusResolution = Float()
```
delarosatrevin: I don't  really see the need to modify the underlying data model (which, by the way, should be in a consensus among different groups developing Scipion) becase a new metric in Xmipp (aka consensusResolution). I think the Data Model in Scipion is quite flexible. So, if Xmipp protocol/program want to set the "consensusResolution" it can do it over a CTFModel object without the need to change the class. Other programs might use it as well if present and react if not there as well.

pyworkflow/em/data.py
```diff
@@ -192,6 +193,9 @@ def hasResolution(self):
     def setResolution(self, value):
         self._resolution.set(value)
 
+    def setConsensusResolution(self, value):
```
delarosatrevin: Same here, I don't like to add more methods to the base Data Model if not really necessary.

pyworkflow/em/packages/xmipp3/protocol_preprocess_micrographs.py
```diff
@@ -310,6 +310,7 @@ def _updateOutputSet(self, outputName, outputSet, state=Set.STREAM_OPEN):
         else:
             # Here the defineOutputs function will call the write() method
             self._defineOutputs(**{outputName: outputSet})
+            self._defineSourceRelation(self.inputMicrographs, outputSet)
```
delarosatrevin: A "transformRelation" (which is also a source relation) should be more appropriated here. I guess at this point the outputSet is the result of a transformation of the input one. This is a more strong relation (one to one mapping in sets) than just source relation.

pyworkflow/em/data.py
pconesa: Yes, @majtner is on changing this. A change in the model is not needed.

pyworkflow/em/packages/xmipp3/protocol_preprocess_micrographs.py
pconesa: to be honest.....the output is == input + 2 new columns......the CTF remains the same. Does it fit under "transform"?

majtner: This file is actually protocol_preprocess_micrographs.py not ctf_consensus, sorry for the confusion.
During the tests, I was using also preprocessing and I noticed that there is missing source relation, so I added it there. It could be easily changed to transformRelation, if it is more suitable here.

delarosatrevin: Yes, it is indeed a transform! Source just mean that you used some input (s) to obtain a result output. Transform is a stronger relation (that implies source relation) and said that the input object was "transformed" into the output object. Of course, there is not a clear line when to use one or another, but I think this case is more a transform.

pconesa:
We can't merge any PR that will affect xmipp code (c++ nor python)....until Xmipp split is done. I'm closing this.

PR #1656, opened 2018-06-15
Adding LocScale method.

dmaluenda:
- Including the protocol to the protocols.template
 - Including the LOCSCALE_HOME in scipion.template
 - Adding the package to the install.script.py
 - Adding the protocol in the new em.package.sachselab
 - Adding the test in the tests.em.protocols

### DIFF
--- a/pyworkflow/em/packages/sachselab/__init__.py
+++ b/pyworkflow/em/packages/sachselab/__init__.py
@@ -1,0 +1,46 @@
+# **************************************************************************
+# *
+# * Authors:     David Maluenda (dmaluenda@cnb.csic.es)
+# *
+# * Unidad de  Bioinformatica of Centro Nacional de Biotecnologia , CSIC
+# *
+# * This program is free software; you can redistribute it and/or modify
+# * it under the terms of the GNU General Public License as published by
+# * the Free Software Foundation; either version 2 of the License, or
+# * (at your option) any later version.
+# *
+# * This program is distributed in the hope that it will be useful,
+# * but WITHOUT ANY WARRANTY; without even the implied warranty of
+# * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# * GNU General Public License for more details.
+# *
+# * You should have received a copy of the GNU General Public License
+# * along with this program; if not, write to the Free Software
+# * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
+# * 02111-1307  USA
+# *
+# *  All comments concerning this program package may be sent to the
+# *  e-mail address 'scipion@cnb.csic.es'
+# *
+# **************************************************************************
+"""
+This package contains the protocols and data for LocScale
+"""
+from locscale import *
+from bibtex import _bibtex # Load bibtex dict with references
+
+_logo = "locscale_logo.jpg"
+
+from protocol_locscale import ProtLocScale
+
+def validateInstallation():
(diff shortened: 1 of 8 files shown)


pyworkflow/em/packages/sachselab/__init__.py
```diff
@@ -0,0 +1,46 @@
+# **************************************************************************
```
delarosatrevin: The developer of localScale (Arjen Jakobi) has not his own group in Delf, so not sure if we want to group this under 'sachselab' (also check proper spelling) that was his previous group.

pyworkflow/em/packages/sachselab/locscale.py
```diff
@@ -0,0 +1,127 @@
...
+
+import os
+
+from pyworkflow.utils import Environ, replaceBaseExt
+from pyworkflow.em.convert import ImageHandler
+
+# we declarate global constants to multiple usage
+LOCSCALE_HOME_VAR = 'LOCSCALE_HOME'
+EMAN2DIR_VAR = 'EMAN2DIR'
+LOCSCALE_HOME = os.environ[LOCSCALE_HOME_VAR]
```
delarosatrevin: It would be good to provide a default value to be more robust to a config file that has not yet defined it.

pyworkflow/em/packages/sachselab/locscale.py
```diff
@@ -0,0 +1,127 @@
...
+        errors.append('Eman (v: %s) is needed to execute this protocol'
+                      % ', '.join(getSupportedEmanVersions()))
+    return errors
+
+def setEmanEnviron():
+    """ Setup the environment variables needed to launch Eman and 
+        use its modules.
+    """
+    env = Environ(os.environ)
+    pathList = [os.path.join(EMAN2DIR, d)
```
delarosatrevin: Here another problem could be that the EMAN2 version installed in Scipion not necessarily  match with the one required by locScale

pyworkflow/em/packages/sachselab/__init__.py
dmaluenda: I will directly ask A. Jakobi how he want to call the package (not the method).

Thanks!

pyworkflow/em/packages/sachselab/locscale.py
dmaluenda: At the beginning I started defining a env.var. LOCSCALE_EMAN_HOME to point to the locscale's required version of Eman, independently of the Eman default version (in the same way of LocalRec and Relion). However, I dismissed that because it finally works with the three current versions of the Eman.

I can return to that point to prevent futures incompatibilities. What do you think?

delarosatrevin: Hum, not sure...it is certainly more clear as it is now. Maybe we can let it be this way and change in the future if needed.

pyworkflow/em/packages/sachselab/locscale.py
dmaluenda: Good point! Pushed.

Thanks!

pyworkflow/em/packages/sachselab/__init__.py
dmaluenda: ’Sachselab’ would be appropriate.

pconesa:
@delarosatrevin should we merge this then?

delarosatrevin:
It looks fine, I haven't had the time to test it properly.

PR #1662, opened 2018-06-21
Fix problem with join set

majtner:
Fixing problem with joining SetOfClasses2D

### DIFF
--- a/pyworkflow/em/protocol/protocol_sets.py
+++ b/pyworkflow/em/protocol/protocol_sets.py
@@ -132,7 +132,11 @@ class ProtUnionSet(ProtSets):
         set1 = self.inputSets[0].get()  # 1st set (we use it many times)
 
         # Read ClassName and create the corresponding EMSet (SetOfParticles...)
-        outputSet = getattr(self, "_create%s" % set1.getClassName())()
+        if set1.getClassName() == "SetOfClasses2D":
+            imgs = set1.getImages()
+            outputSet = getattr(self, "_create%s" % set1.getClassName())(imgs)
+        else:
+            outputSet = getattr(self, "_create%s" % set1.getClassName())()
 
         # Copy info from input sets (sampling rate, etc).
         outputSet.copyInfo(set1)  # all sets must have the same info as set1!


josuegbl:
Have you create a test to join classes 2D? Anyway, I'm not sure if the changes that you are proposing is the right way to manage joins of SetOfClasses.... I think you should also join the SetOfparticles that are relates with these SetOfClasses (either 2D or 3D). Then, make sure the ids of the particles are properly corrected (if you merge two or more set that contains particles, id conflict could be occurs.)

majtner:
No, I did not create a test. I simply needed to join more sets of classes and this protocol was failing for SetOfClasses2D (despite proposing that it can merge "all" sets), so I fixed it for my purposes.

I do not think that merging also SetOfParticles is feasible. In case you have multiple classifiers producing multiple outputs with different distribution of particles between classes, how do you want to proceed? Even in case you localize "duplicate" particles, which are also used in some different class, after removing them you want to re-calculate the averages? In extreme case you can even completely remove some input classes (because all images will be localized in some other classes from some other classifiers) and I do not think this additional functionality is what user expects when he/she executes joining sets.

josuegbl:
Well I think you should create a test, and I also think the fixes should works for the general purposes, not only for yours.
Merge the SetOfParticles(SOP) is a "must", because every SetOfClasses (SOC) is associated a SOP. So, if you join several SOC, you are internally merging particles coming from different SOP. If you dont the merge, which is your SOP associated to your new SOC, that contains ALL the particles that are in your new SOC?
In the case of the duplicated IDs (not PARTICLES, just Ids) you MUST change its Ids, because in scipion its one of the more important properties.  In the case when you have the SAME particle duplicated, you must remove one. So, I think that the user expects that the protocols works correctly and not support the same particle duplicated in the same set, which is a big error.

majtner:
Ok, I sincerely believed that this change could help others (this is why I created it, I really do not need it to work only for me, you can trust me on this ;-) ). 
The protocol is now failing but if you think it is better than this change, you can close it, I have no problem with it.

pconesa:
Thanks both (@majtner and @josuegbl ) I'll meet Tomas on Monday and try to clarify this taking into account Josue's feedback. I still don't understand, how a merge SOC (I like this new naming concept) as proposed here is of any use. But Hope to clarify this with @majtner, if it make sense, somehow I vote to keep it...I it doesn't, I vote to exclude SOC.

delarosatrevin:
@pconesa Maybe we can discuss it on Tuesday. I agree with the concerns raised by @josuegbl. Merging the SetOfClasses is not that easy to handle. 
1) As he said, each SetOfClasses is associated with a input Set, so, the merge of the input particles should be done
2) The duplication of IDs is more tricky in this case

I would said that maybe we need to discuss the scenarios  in which make sense to merge classes...if not, maybe it is better to prevent users to merge this data type.

majtner:
I think that the merging makes more sense for SetOfAverages, which is already listed as Input type. Since the output of classification is typically SetOfClasses2D, it was straightforward to merge it directly without conversion to SetOfAverages but as it is obvious from the discussion, it is not so easy.

Based on my view, the main problem is that Input type allows "All" but it is not working for all sets and probably also does not make sense for all sets.
The easiest solution seems to be excluding option "All" and keep merging only for those sets, which are explicitly listed, currently Particles, Micrographs, CTFs, Volumes, Averages (maybe we can add more if necessary).

delarosatrevin:
You are right @majtner, maybe we should remove the "All" option there. It is True that some of the Set objects are difficult to merge. For example, SetOfCoordinates has the same problem as SetOfClasses, it references an input Set and then the merge get too much complicated. 

So, I like the idea of disabling 'All' option and maybe avoid some sets to be merged right now.

pconesa:
I've created the issue and I'm closing this. Honestly....many thanks to all 3!! For your help.
#1664

PR #1665, opened 2018-06-25
Model 09 emringer

mmmtnez:
Dear all, 
here you are the model building worflow updated with two main points:

1.- To adapt the CCP4 environment to the new CCP4 package version (7.0), we have setup the CCP4 convert protocol (pyworkflow/em/convert_header/CCP4/convert.py) by splitting the variables; some of them will be added, whereas others will be replaced.

2.- We have added the PHENIX package (version 1.13) to the pyworkflow/em/packages/ path to introduce two validation protocols, EMRinger and Molprobity. In this pull request we are going to add the first one, EMRinger. The second validation protocol, Molprobity, will be added in the next pull request in the coming days (commented line in pyworkflow/em/packages/phenix/__init__.py). By now, we have created the variables of PHENIX environment, as well as the requirements to run phenix and those mentioned programs (pyworkflow/em/packages/phenix/convert.py).

Concerning EMRinger validation protocol:

This protocol validates an atomic structure (PDBx/mmCIF) against its electron density map. The inputs are thus the volume and the atomic structure. After running phenix EMRinger a summary of validation numerical values and several files are generated, and we provide a visualization panel to check those values and files in a similar way as PHENIX does. In addition, the summary values are also showed in the SCIPION window interface. To capture all these values, as well as the information of structure chains and rotameric values of individual residues, we inspect the .pkl files generated by phenix python. We have accomplished this inspection by using also phenix.python to take advantage of variables and libraries, and to avoid problems with the python version. The captured values are written in a text file that, after being transformed in .py file, can be read by our scipion python in the common way. The text and .py files are temporarily saved, whereas the common files generated by phenix EMRinger are saved in the extra folder.

To test the appropriate performance of the EMRinger protocol, we show the use of this program in the common workflow of model building, directly after importing a volume and an atomic structure or after running the refinement programs coot or refmac. Remark that ONLY FOR TESTING PURPOSES, and in order to check EMRinger results showed in the text file, the temporarily files are saved in the extra folder.

scipion test tests.em.workflows.test_workflow_modeling.TestEMRingerValidation


config/templates/scipion.template
```diff
@@ -23,7 +23,9 @@ BSOFT_HOME = %(EM_ROOT)s/bsoft-1.9.0
 CRYOEF_HOME = %(EM_ROOT)s/cryoEF-1.1.0
 CTFFIND_HOME = %(EM_ROOT)s/ctffind-3.6
 CTFFIND4_HOME = %(EM_ROOT)s/ctffind4-4.1.10
-CCP4_HOME =  %(EM_ROOT)s/ccp4-7.0
+CCP4_HOME =  %(EM_ROOT)s/ccp4
```
josuegbl: This not follow the convention in Scipion (PACKAGE_HOME = %(EM_ROOT)s/PAKCAGE-version)

pyworkflow/em/convert_header/CCP4/convert.py
```diff
@@ -43,15 +43,24 @@
 START = 1  # save coordinate origin in the mrc header field=start (pixel)
```
josuegbl: I think here are mixing some general functions that could be on utils and specific functions that should be into the package folder. I think create a new general folder (convert_header) that integrates several dependencies with one (o more packages) is a very bad idea, and it goes in the opposite direction of the pluginization.  I strongly recommend review these and split the general and specific functions putting in the right place.

pyworkflow/em/data.py
```diff
@@ -610,6 +610,10 @@ def getVolOriginAsTuple(self):
         return x, y, z
         # x, y, z are floats in Angstroms
 
+    def setVolOriginAsTuple(self, x, y, z):
```
josuegbl: I think you are duplicating functions here. You can treat the origin as a tuple in the original function (setOrigin), and not create another one to do that. The functions should be only one, independently to the format of the data. I think these is a bad coding practice

pyworkflow/em/packages/xmipp3/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -389,8 +389,21 @@ def _postprocessOutput(self, volumes):
...
+                volumes.setSamplingRate(self.samplingRate)
+                # we have a new sampling so origin need to be adjusted
+                iSampling = self.inputVolumes.get().getSamplingRate()
+                oSampling = self.samplingRate
+                xdim_i, ydim_i, zdim_i = self.inputVolumes.get().getDim()
+                xdim_o, ydim_o, zdim_o = volumes.getDim()
+
+                xOrig, yOrig , zOrig = \
+                    self.inputVolumes.get().getVolOriginAsTuple()
+                xOrig += (xdim_i*iSampling-xdim_o*oSampling)/2.
```
josuegbl: I think the way to estimate the new origin is wrong. Test case: a volume resize form 200 to 100 (pixel size 1), so initial origin 100, xdim_i 100, xdim_o 50,  xOrig += (100 * 1-50 * 1)/2 = 125 when should be 50.

mmmtnez:
We have consider and modify the files according your suggestions, changes have been included in the new pull request model_09_emringer_updated. So, ignore and close this pull request.

delarosatrevin:
Should this PR be closed with the new "Model 09 emringer updated" PR???

PR #1680, opened 2018-07-03
Pluginization install config

yaizar:
Merge pluginization branch. Folders bsoft and grigoriefflab will disappear from pw/em/packages.


azazellochg:
@yaizar so, what would be a testing strategy? Remove vars from ~/.config/scipin/scipion.conf, remove all grigoriefflab binaries from software/em and try to play with bsoft/grigoriefflab plugins?

pconesa:
That would do...even you should be able to discover test and run them.

azazellochg:
First issue: scipion install --help prints normal output but then starts to compile xmipp

delarosatrevin:
Two comments from my here:
# Why not to group the plugin-related templates under a folder that could be copied and used as starting point? 
# I think all the plugin and plugin-binaries modes should not be added to the main scipion script. This one is more high level. Maybe all those modes should be merged under 'plugin' mode and then it will contains subcommands for each task (as it is done in other cases, e.g. testdata, etc)

pconesa:
Guys! @yaizar @azazellochg , @delarosatrevin , @dmaluenda , @ldelcano we have clone this branch to I2PC. Please use that one from now. @delarosatrevin had some suggestions and, and are going to be incorporated into the pluginization. Already done plugins might need some minor Refactorization, but it will be cleaner.

New branch is this one: https://github.com/I2PC/scipion/tree/pluginization_install_config

delarosatrevin:
Thanks @pconesa. 

I have talked to @azazellochg and @josuegbl and we agreed on "freeze" on devel the packages that we are "maintainers". So I will start with those for testing the new changes in the pluginization.

https://docs.google.com/document/d/1U9RrmeooYDrrZ7xT5U7VMEZZR2THBnbA7S7z1mET78c/edit?usp=sharing

PR #1681, opened 2018-07-04
MovieMaxShift for release-1.2.1

dmaluenda:
This protocol reject movies taking certain criteria (chosen by the user) related with the shifts evaluated in a previous motion correction protocol. The protocol returns a set of accepted movies and one for the rejected ones. Additionally, if the setOfMovies has a setOfMicrographs associated, the preotocol also returns a set of accepted micrographs and one for rejected ones. The protocol work fine in streaming.

 - Including the protocol to the protocols.template
 - Adding the protocol to the xmipp packages
 - Adding the protocol to the movie alignment targets
 - Adding the test in the tests.em.protocols.test_protocols_xmipp_movies
 - Reviewing the alignment plotter and correcting the bug (shifts are absolutes to certain origin, however it was plotting as a increments between frames)

### DIFF
--- a/config/templates/protocols.template
+++ b/config/templates/protocols.template
@@ -21,6 +21,7 @@ Protocols SPA = [
 	{"tag": "section", "text": "Movies", "openItem": "False", "children": [
 		{"tag": "protocol", "value": "XmippProtOFAlignment", "text": "default"},
 		{"tag": "protocol", "value": "XmippProtMovieCorr", "text": "default"},
+		{"tag": "protocol", "value": "XmippProtMovieMaxShift", "text": "default"},
 		{"tag": "protocol", "value": "XmippProtMovieAverage", "text": "default"},
 		{"tag": "protocol", "value": "ProtUnblur", "text": "default"},
 		{"tag": "protocol", "value": "ProtSummovie", "text": "default"},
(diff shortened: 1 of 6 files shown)


delarosatrevin:
@dmaluenda This branch is frozen for release, so I'm not sure if would be appropriated to merge this PR here. Moreover, I'm not sure if it is necessary to generate extra setOfMicrograph when the user can easily create a subset, but this is just a suggestion, it is up to you as the developer the final decision.

pconesa:
@delarosatrevin ...in an automated pipeline, having a set of micrographs is needed.

dmaluenda:
I know that this branch is frozen, however I think it is a very confined code that no affect others and is a tool for the streaming procedure.

In addition, the bug related to the alignment plotter is quite serious... we are reporting wrong info...

For that reason I considered to do the PR. Nevertheless, if we consider that is not able to merge, I agree.

PD: sorry, I had the intention to discuss with all you the including this to the 1.2.1 but I move to another work and I forget it...

josuegbl:
@dmaluenda, I propose you split the code and do 2 PR: the first one only with the changes related to the plotter and the last one things xmipp related. I agree with @delarosatrevin that all xmipp code (c++ and python) are frozen, so all PRs containing xmipp code should be wait until the decouple.

dmaluenda:
I think that release-1.2.1 branch is frozen because it will be released soon. But the xmipp code (python and c++) is not frozen here, because the scipion/xmipp decoupling is doing in devel branch -not here-. Isn't it??

I agree that it may be splited in two PR (ploter and protocol), however both are very related... as you want. But first, let's discuss whether we want to include this to the v1
2.1.

Thanks to all you for your comments!

josuegbl:
@dmaluenda, I think if you introduce now these changes in to release-1.2.1, several conflicts will be arise when you merge with devel, basically cause all xmipp code is now separated (on a new repository!!!). I vote "No" to merge this PR into release-1.2.1, first because is a frozen code and these is interfering with the separation of Xmipp repositories.

PR #1688, opened 2018-07-10
Model 09 emringer updated

mmmtnez:
Dear all,
in the following text you have the old text of closed pull request model_09_emringer. At the end you will find the answers to the suggestions made to that pull request:
********************************************************************************************************************************
Dear all,
here you are the model building worflow updated with two main points:

1.- To adapt the CCP4 environment to the new CCP4 package version (7.0), we have setup the CCP4 convert protocol (pyworkflow/em/convert_header/CCP4/convert.py) by splitting the variables; some of them will be added, whereas others will be replaced.

2.- We have added the PHENIX package (version 1.13) to the pyworkflow/em/packages/ path to introduce two validation protocols, EMRinger and Molprobity. In this pull request we are going to add the first one, EMRinger. The second validation protocol, Molprobity, will be added in the next pull request in the coming days (commented line in pyworkflow/em/packages/phenix/init.py). By now, we have created the variables of PHENIX environment, as well as the requirements to run phenix and those mentioned programs (pyworkflow/em/packages/phenix/convert.py).

Concerning EMRinger validation protocol:

This protocol validates an atomic structure (PDBx/mmCIF) against its electron density map. The inputs are thus the volume and the atomic structure. After running phenix EMRinger a summary of validation numerical values and several files are generated, and we provide a visualization panel to check those values and files in a similar way as PHENIX does. In addition, the summary values are also showed in the SCIPION window interface. To capture all these values, as well as the information of structure chains and rotameric values of individual residues, we inspect the .pkl files generated by phenix python. We have accomplished this inspection by using also phenix.python to take advantage of variables and libraries, and to avoid problems with the python version. The captured values are written in a text file that, after being transformed in .py file, can be read by our scipion python in the common way. The text and .py files are temporarily saved, whereas the common files generated by phenix EMRinger are saved in the extra folder.

To test the appropriate performance of the EMRinger protocol, we show the use of this program in the common workflow of model building, directly after importing a volume and an atomic structure or after running the refinement programs coot or refmac. Remark that ONLY FOR TESTING PURPOSES, and in order to check EMRinger results showed in the text file, the temporarily files are saved in the extra folder.

scipion test tests.em.workflows.test_workflow_modeling.TestEMRingerValidation

********************************************************************************************************************************
According to the suggestions made for pull request model_09_emringer we have made the next modifications:

1.- config/templates/scipion.template 
Josue wrote: This not follow the convention in Scipion (PACKAGE_HOME = %(EM_ROOT)s/PAKCAGE-version)
Answer: We have included the version of CCP4 and PHENIX packages

2.- pyworkflow/em/convert_header/CCP4/convert.py 
Josue wrote: I think here are mixing some general functions that could be on utils and specific functions that should be into the package folder. I think create a new general folder (convert_header) that integrates several dependencies with one (o more packages) is a very bad idea, and it goes in the opposite direction of the pluginization. I strongly recommend review these and split the general and specific functions putting in the right place.
Answer: We have split the content of this file in two new ones. The first one (pyworkflow/em/convert_header/CCP4/convert.py) preserves every methods useful for converting to CCP4/mrc volumes of other formats, that can be used in different protocols of chimera or phenix packages, for example. The second one (pyworkflow/em/packages/ccp4/convert.py) includes methods required to get the CCP4 environment or to access to CCP4 programs.

We still have to consider the last two suggestions of model_09_emringer pull request


josuegbl:
@mmmtnez, @rmarabini and @pconesa, why these changes are not included in the previous PR instead of create a new one? I think this practice twice the efforts to review because I need re-review all the code instead of only review the changes...

mmmtnez:
You're right, but there were some last additional protocols that we have decided not to include in this pull request and let them for the next one. Since removing this last part of the pull request seemed more difficult than create another one, we have decided to submit the old pull request with this new name (model_09_emringer_updated). Anyway, you do not have to re-review all the code because only the suggested changes have been accomplished. Consider that this pull request is in fact the old one although with a different name.

pconesa:
@josuegbl, consider this PR an exact copy of the previous one.....that will also contain your suggestions.

pyworkflow/em/data.py
```diff
@@ -610,6 +610,10 @@ def getVolOriginAsTuple(self):
         return x, y, z
         # x, y, z are floats in Angstroms
 
+    def setVolOriginAsTuple(self, x, y, z):
```
josuegbl: PLease, see my comment in previous PR #1665

pyworkflow/em/packages/xmipp3/protocol_preprocess/protocol_crop_resize.py
```diff
@@ -389,8 +389,21 @@ def _postprocessOutput(self, volumes):
...
             if self.doResize:
-                volumes.setSamplingRate(self.samplingRate)            
-     
+                volumes.setSamplingRate(self.samplingRate)
+                # we have a new sampling so origin need to be adjusted
+                iSampling = self.inputVolumes.get().getSamplingRate()
+                oSampling = self.samplingRate
+                xdim_i, ydim_i, zdim_i = self.inputVolumes.get().getDim()
+                xdim_o, ydim_o, zdim_o = volumes.getDim()
+
```
josuegbl: PLease, see my comment in previous PR #1665

pyworkflow/em/convert_header/CCP4/convert.py
```diff
@@ -31,107 +31,14 @@
 """
```
josuegbl: @mmmtnez, regarding your answer about these, I still believing that these creates a dependency to CCP4 package. Also, you must use the proper functions (instead of creates new duplicated) that we are developed to convert formats in Scipion (please, see pyworkflow/em/convert.py). I think all files/directories should be removed from convert_header.

josuegbl:
@pconesa, I needed re-review all the PR again because there are several recommendations that has not been addressed. So, I needed to treat these PR as a new one and not as "exact copy".... Please, let's make an effort not to do the work twice

pyworkflow/em/data.py
mmmtnez: Original Comment:
"I think you are duplicating functions here. You can treat the origin as a tuple in the original function (setOrigin), and not create another one to do that. The functions should be only one, independently to the format of the data. I think these is a bad coding practice"

Certainly, setVolOriginAsTuple is not a good name so we have renamed it as setShiftInOrigin. The difference between setOrigin and setShiftInOrigin is that the first updates the whole origin (a 4x4 matrix) while the second  assumes that there are no rotations and only updates a shift vector.

pyworkflow/em/packages/xmipp3/protocol_preprocess/protocol_crop_resize.py
mmmtnez: Original question: "I think the way to estimate the new origin is wrong. Test case: a volume resize form 200 to 100 (pixel size 1), so initial origin 100, xdim_i 100, xdim_o 50, xOrig += (100 * 1-50 * 1)/2 = 125 when should be 50."

Origin is stored in Angstroms. Therefore, if you scale an image the origin should NOT change. Certainly, in pixels, it does change but not in  Angstroms.

So if we change the sampling rate from 1 A/px to 2 A/px xDim will change from 200px to 100px and therefore
 +                xOrig += (xdim_i*iSampling-xdim_o*oSampling)/2. 
                   100   += (200*1 -100*2)/2
                   100 +=0
                   newxOrig = oldXorig = 100 A.

The situation changes if we crop instead of scale.

pyworkflow/em/convert_header/CCP4/convert.py
mmmtnez: pyworkflow/em/convert_header/CCP4/convert.py  does not depend on any protocol based on CCP4 programs so we are not sure what is the meaning of "dependency to CCP4 package". The objective of pyworkflow/em/convert_header/CCP4/convert.py  is to manipulate the CCP4 header which is interesting for programs in several different packages. Actually the name convert is misleading, no convert is done here, so we should rename from convert_header to handle_header.

Certainly, the access to different image header formats may be centralize in a single place. I do not think the right place for pyworkflow/em/convert_header/CCP4/convert.py is   em/convert.py since the aim of the class is not to convert between headers or provide  transparent access to any header but to deal with CCP4 header. In the future,  somebody may be able to create a class similar to imageHandler for the headers but we are not there. We are just building the reading/writing utility for one of the formats.  

In summary, we may rename convert_header to handle_header, people who develop routines to access different headers may add them in this directory and eventually a general headerHandler may be created and integrated into imageHandler

mmmtnez:
Hi @pconesa, @delarosatrevin  and @josuegbl ,

  We would appreciate a second opinion about a comment made by @josuegbl.

   As far as we know we have followed or answered all Josue's comments. The only one we have only partially follow is 

> "@mmmtnez, regarding your answer about these, I still believing that these creates a dependency to CCP4 package. Also, you must use the proper functions (instead of creates new duplicated) that we are developed to convert formats in Scipion (please, see pyworkflow/em/convert.py). I think all files/directories should be removed from convert_header."

our point of view is:

> 
> 
> 
> pyworkflow/em/convert_header/CCP4/convert.py does not depend on any protocol based on CCP4 programs so we are not sure what is the meaning of "dependency to CCP4 package". The objective of pyworkflow/em/convert_header/CCP4/convert.py is to manipulate the CCP4 header which is interesting for programs in several different packages. Actually the name convert is misleading, no convert is done here, so we should rename from convert_header to handle_header.
> 
> Certainly, the access to different image header formats may be centralize in a single place. I do not think the right place for pyworkflow/em/convert_header/CCP4/convert.py is em/convert.py since the aim of the class is not to convert between headers or provide transparent access to any header but to deal with CCP4 header. In the future, somebody may be able to create a class similar to imageHandler for the headers but we are not there. We are just building the reading/writing utility for one of the formats. In a sense this class is equivalent to the c file ./software/em/xmipp/libraries/data/rwMRC.cpp
>
>
> In summary, we may rename convert_header to handle_header, people who develop routines to access different headers may add them under this directory and eventually a general headerHandler may be created and integrated into imageHandler
>

pconesa:
Well, I see your point @mmmtnez , @rmarabini ....If I understand correctly this is like a "generic utils" that you see it will be useful for any "modelling package" and therefore does not strictly belong to a package.

In a sense, could it be considered similar to file format conversion methods that do not belong (or shouldn't) to any package?

For me, it is clear it should be outside the package and there is no "digital dependency to CCP4 package. Whether if should be placed inside the already existing convert.py or deserves a new space....is a matter of how do we organize things.....do not have any strong opinion about any of the options here. If a growth in functionality is expected in this "tools" I rather keep it separate and leave it as it is.

pyworkflow/em/convert_header/CCP4/convert.py
josuegbl: Well @mmmtnez, or could I name you @rmarabini? :tongue: I see your point, and I think you are right, at least in part :smile: ... I think the chosen name for the file (and also the directory) is inconvenient and lead to misunderstanding. I agree rename convert_header to handle_header but also rename CCP4/convert.py to header.py (removing the directory CCP4) or header_convert.py or header_handler.py: I'm open with the name of the .py file.

pyworkflow/em/packages/xmipp3/protocol_preprocess/protocol_crop_resize.py
josuegbl: OK to me. You are right.

pyworkflow/em/data.py
```diff
@@ -602,14 +602,18 @@ def _getDefaultOrigin(self):
...
-    def getVolOriginAsTuple(self):
+    def getShiftsFromOrigin(self):
         origin = self.getOrigin(force=True).getShifts()
         x = origin[0]
         y = origin[1]
         z = origin[2]
         return x, y, z
         # x, y, z are floats in Angstroms
 
+    def setShiftsInOrigin(self, x, y, z):
```
josuegbl: It's better now. I'm still believing you are killing flies with nuclear bombs. My point is if you are using Transform class to store the origin, all information about angles (euler  3x3 matrix) is totally unnecessary cause an origin only defines a centre of references, and is invariant to angles (it's a point, not a vector). Anyway, I can still living with this.

pyworkflow/em/protocol/protocol_import/volumes.py
```diff
@@ -244,15 +244,17 @@ def _defineParams(self, form):
...
-            pdbPath = self._getPath('%s.cif' % self.pdbId.get())
+            pdbPath = self._getTmpPath('%s.cif' % self.pdbId.get())
             self._insertFunctionStep('pdbDownloadStep', pdbPath)
         else:
             pdbPath = self.pdbFile.get()
         self._insertFunctionStep('createOutputStep', pdbPath)
 
     def pdbDownloadStep(self, pdbPath):
         """Download all pdb files in file_list and unzip them."""
+        print "AAAAAAAAAAAAAAAAAAAAAAAAAAAAA: ", pdbPath
```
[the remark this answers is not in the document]
mmmtnez: Done

pyworkflow/em/convert_header/CCP4/convert.py
delarosatrevin: Dear all, following is my "second opinion" about this.

I agree that if it is a general enough functionality and does not depend on anything from CCP4, it can be in that place and not specifically for ccp4 package. 

Nonetheless, I don't like the name convert_header (or handle_header)/CCP4/convert.py
I will vote for something shorter and more generic, for example, headers.py and put the functions/classes there. This will allow to import the classes as:

from pyworkflow.em.headers import Ccp4Header, etc

Later, if the headers functionality needs to be extended, it can be included in the same file or create a directory for the submodule such as:

headers
     __init_.py  -> from ccp4 import *; from xxx import *; etc
     ccp4.py
     xxx.py

and the imports from the outside will still be accessible in the same way.
from pyworkflow.em.headers import Ccp4Header

mmmtnez:
We have renamed  convert_header/CCP4/convert.py to headers.py. We think this was the only pending request so this is ready to go.

PR #1692, opened 2018-07-16
Release 1.2.1 dm4

pconesa:
Enable Scipion/xmipp to read mrc4 bits and visualise them
Handle Java max memory (2Gb) by default +1 (3GB) for set of movies. 
If JAVA_MAX_MEMORY in the config it will be taken into account.


pconesa:
@delarosatrevin , maybe we want to merge this straight away and do the PR to release-1.2.1 instead?

PR #1693, opened 2018-07-17
Release 1.2.1 dm4

pconesa:


pyworkflow/em/showj.py
```diff
@@ -103,6 +103,11 @@
 OBJECT_ID = 'objectId'
 
 
+def getJVMMaxMemory():
```
delarosatrevin: This does not follows our convention...for readability it should be:

getJvmMaxMemory

pconesa:
Renamed!

Issue #1043, opened 2017-06-05
unify label


rmarabini:
Currently we have three protocols called

   consensus coordiantes
   classes consensus
   ctf comparison

the three should have a more consistent name as

   coordiantes  consensus
   classes consensus
   ctf  consensus

I volunteer to do it

   Roberto

delarosatrevin:
You are a very clever guy @rmarabini, this is an easy one! Still someone has to do it. I agree with the protocol labels change for consistency. 

Coul you also take a look at "assign" ones? See the attached image. I propose that all start with the verb: 
assign XXXX
<img width="499" alt="screen shot 2017-06-05 at 17 40 07" src="https://cloud.githubusercontent.com/assets/8441688/26791033/269022ba-4a16-11e7-9891-5f6ecc7ce569.png">

rmarabini:
OK.

delarosatrevin:
@rmarabini, Have you addressed this? Can we close the issue?

rmarabini:
NO, I did not look to this one


   Roberto

PR #1695, opened 2018-07-18
Unify label. Solving #1043

the-best-elephant:
Unify labels: named *consensus or *assign


delarosatrevin:
Although these are quite simple changes, can it wait until the pluginization?

pconesa:
It could...but remember we still have changes to come from release-1.2.1.

PR #1706, opened 2018-07-30
sharpening for release1.2.1

erneyramirez:
A new protocol for sharpening.


pyworkflow/em/packages/xmipp3/protocol_volume_local_sharpening.py
```diff
@@ -0,0 +1,277 @@
...
+import numpy as np
+import os
+import pyworkflow.em.metadata as md
+from pyworkflow.em.metadata.constants import (MDL_COST, MDL_ITER, MDL_SCALE)
+from ntpath import dirname
+from os.path import exists, lexists
+
+LOCALDEBLUR_METHOD_URL= 'http://github.com/I2PC/scipion/wiki/XmippProtLocSharp'
+
+CHIMERA_RESOLUTION_VOL = 'MG_Chimera_resolution.vol'
```
azazellochg: I think it's better to use fileDict for these things. [Here](https://github.com/I2PC/scipion/blob/master/pyworkflow/em/packages/relion/protocol_base.py#L86) is an example

erneyramirez:
Dictionary Done!
Thank Gregory

pyworkflow/em/packages/xmipp3/protocol_volume_local_sharpening.py
```diff
@@ -0,0 +1,278 @@
...
+
+        form.addParam('resolutionVolume', PointerParam, pointerClass='Volume',
+                      label="Resolution Map", important=True,
+                      help='Select a local resolution map.')
+                
+        form.addParam('const', FloatParam, default=1, 
+                      expertLevel=LEVEL_ADVANCED,
+                      label="lambda",
+                      help='Regularization Param.' 
+                        'The software determines this parameter automatically.'
```
josuegbl: I think you can improve much more this help. Which is the meaning of this parameter? In which cases is necessary change the default values? The help is really to help the users to properly use your methods.

pyworkflow/em/packages/xmipp3/protocol_volume_local_sharpening.py
```diff
@@ -0,0 +1,278 @@
...
+                         'However, if the user wishes it can be modified')
+        form.addParam('K', FloatParam, default=0.025, 
+                      expertLevel=LEVEL_ADVANCED,
+                      label="K",
+                      help='K = 0.025 works well for all tested cases'
```
josuegbl: same here... more detailed help is more than welcome.

erneyramirez: In principle, the idea is not to modify these parameters. I have put them for some cases that may give problems. Perhaps for that it is better to read the article. In any way, I will try to improve it.

josuegbl: Although the user should not modify these parameters in most of cases, I think a better explanation on the meaning of these parameters is necessary for the well understanding of your method without consulting every time the paper. You can see others packages (like Relion) as example.

josuegbl:
I found a bug when use resolution maps from relion or resmap: #1713

delarosatrevin:
Some extra minor comments:
There are a few cases where abc + "/" + xyz can be replaced by a neat Python function: os.path.join(abc, xyz)

erneyramirez:
Ok Jose Miguel!
Josue, I've tried to improve the help. Anything that can be improved, you tell me.
Resmap, Relion resolution map, is not exactly a bug, rather it is not made for them. In any case, I put it in the help for some interested user. we will surely expand it in the future.

erneyramirez:
I have made small changes in the protocol so that resmap and blocres maps can be used

josuegbl:
Very good news @erneyramirez!! The help is better (although Im sure you can improve it more 😄). Is there any reason why you not support Relion's resolution map? I think will be better support all resolution maps. Also, maybe we need to think about introduce a new object in Scipion: ReesolutionVolume

delarosatrevin:
The test seems to run fine, but one of the validations is failing:
  File "/home/josem/work/development/scipion/software/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/josem/work/development/scipion/pyworkflow/tests/em/protocols/test_protocols_localdeblur.py", line 106, in testLocalDeblur
    "LocalDeblur  has failed")
  File "/home/josem/work/development/scipion/software/lib/python2.7/unittest/case.py", line 422, in assertTrue
    raise self.failureException(msg)
AssertionError: LocalDeblur  has failed
which correspond to line:
```
        self.assertTrue(exists(LocalDeblur._getExtraPath('sharpenedMap.mrc')),
                        "LocalDeblur  has failed")
```


Maybe the result file names have changed? The following is under the 'extra' folder:

image

PR #1714, opened 2018-08-01
Make Xmipp installation through the plugin.

pconesa:
Make plugin installation params "friendlier" (shorter ans no _)
plugin installation scripts uses new plugin arch (Domain...)
New needed functions for xmipp ghost (for accessing xmipp plugin)
Domain: Calls defineVariables at registering time.


delarosatrevin:
Hi @pconesa 
Even if shorter names, this still adds new 4 modes to the main scipion script. What about the following proposition:
* Keep only two modes: install and uninstall
* scipion (install |  uninstall) (--plugin|--binary) targetName
With this syntax one should be able to install/uninstall plugins and binaries provided by plugins. 
Then, to list available binaries/versions provided by current plugins:
* scipion install --list
This could list current available binaries (grouped by the plugin that provides them)

Even the case of install/uninstall --plugin, it will be redundant, since all plugins will be Python modules, I will argue here just to use the standard method provided by the Python environment being used:
* pip (install / uninstall) ...
* conda (install /uninstall) ..
* etc

In that way we could even reduce the install/uninstall commands for the binaries provided by the plugins.

pconesa:
I agree to refactor the commands...but not in this PR, please. Thanks for having a look @delarosatrevin

PR #1715, opened 2018-08-02
Leaving software/em/xmipp from this repository

dmaluenda:
- Setting the variables for the new bindings from the xmipp repositories.
 - Deleting the whole software/em/xmipp from this repository.
 - Removing the exception !software/em/xmipp from .gitignore.

Now, the xmipp binaries are compiled and installed in software/em with the plugin's installation method. Alternatively (usually for developers), xmipp can be compiled and installed in its own (via xmipp script) while a symbolic link points from xmipp/build to software/em/xmipp-3.18.05.


delarosatrevin:
On top of this change, I will strongly suggest that we remove xmipp history from this repo as well...this make the repository lighter to download. Anyway, it is not recommended to re-write the history in most cases...but we can do it if we move the scipion repo to scipion-em and have the "cleaned" one there. 
What do you think @pconesa, @dmaluenda , @azazellochg , @josuegbl ???

josuegbl:
I agree with you (without it serving as precedent :smile: )

dmaluenda:
We migrated the xmipp C++ sources (and bindings) without keepinh the history alive. Then, after a hard work spliting the sources in core, viz and reconstraction repos, we realized that the history was lost. However, we desestimated to repeat the work.

In summary, xmipp history will be lost... If you still thinking to lighter the repository, with pablo and peter we was talking about keeping a clone of the whole repo as a backup either in our server or in gitHub.

delarosatrevin:
I think we can create a new scipion repo with the lighter history, basically without the xmipp history inside. I think this was something should  be took into account when splitting the xmipp repos, not only because of the code history, but also because of authors contributions...but it wasn't done that way. Anyway, I think there are some (relatively) big files in the history that will be better to get rid of them and make it lighter in the case of Scipion.

PR #1716, opened 2018-08-03
str() added to integer used in print line

pconesa:
upss.


delarosatrevin:
This fix is not longer needed, should we close this PR?

pconesa:
Not needed since it's in other PR.

PR #1721, opened 2018-08-07
xmipp -> xmippLib binding module & import modules from plugins

dmaluenda:
Sorry for mix two different thing in one PR...

- [x] Replacing the `import xmipp` by `import xmippLib` for the binding module (see https://github.com/I2PC/scipion-em-xmipp/pull/1)

- [x]  Introducing a new util importFromPlugin(module, method='', errorMsg='', doRaise=False) and it's uses.

This method try to import the method from the plugin/module and returns what is imported if not fails. When the importation fails (due to the plugin is not found), it prints a common message + optional errorMsg or it raise with the same message if doRaise is True. By default, it don't raise to avoid errors in the very beginning importing loads that crash the scipion launching.


pyworkflow/em/protocol/protocol_import/particles.py
```diff
@@ -153,16 +154,22 @@ def _insertAllSteps(self):
...
         elif self.importFrom == self.IMPORT_FROM_XMIPP3:
-            from pyworkflow.em.packages.xmipp3.dataimport import XmippImport
+            XmippImport = importFromPlugin('xmipp3.convert', 'XmippImport',
+                                           'Xmipp is needed to import .xmd files',
+                                           doRaise=True)
             self.importFilePath = self.mdFile.get('').strip()
             return XmippImport(self, self.mdFile.get())
         elif self.importFrom == self.IMPORT_FROM_RELION:
-            from pyworkflow.em.packages.relion.dataimport import RelionImport
+            RelionImport = importFromPlugin('relion.dataimport', 'RelionImport',
```
delarosatrevin: this should be 'relion.convert'

pyworkflow/em/protocol/protocol_import/particles.py
```diff
@@ -171,7 +178,10 @@ def getImportClass(self):
             return ScipionImport(self, self.importFilePath)    
         elif self.importFrom == self.IMPORT_FROM_FREALIGN:
             self.importFilePath = self.parFile.get('').strip()
-            from pyworkflow.em.packages.grigoriefflab.dataimport import GrigorieffLabImportParticles
+            GrigorieffLabImportParticles = importFromPlugin(
+                     'grigoriefflab.dataimport', 'GrigorieffLabImportParticles',
```
delarosatrevin: 'grigoriefflab.convert'

dmaluenda:
@delarosatrevin, could you have a look on it? please

The external imports (from some plugin) is giving error and all we are trying to fix it in our single scope, whereas here we have all (I think all...) fixed.

dmaluenda:
What do you mean?? You mean to don't use the function??

PD: I've just seen your comment, we commented almost at the same time ;p

PR #1724, opened 2018-08-10
CRITICAL visualization BUG in release-1.2.1

delarosatrevin:
There is a very important bug introduced in #1693 that stop visuazliations to show up mainly for coordinates and some other cases. 

@pconesa and @rmarabini , I think you should have been more careful with the changes introduced in the Java memory param, it was changed from integer to string, but all the cases were not tested. It was not even basically tested that most common showj use cases were working!!! I think we should avoid this issues, specially for a release that is about to be out. You "fixed" the case for visualizing movies, that is not commonly used...and you broke the Xmipp picker, coordinates visualization and other cases.

I will take also my part of responsibility, since I reviewed this PR and this issue passed. The other problem, is this original PR was about DM4 support, which I tested...and this "minor change" in Java viewer was also introduced here. We should try to avoid this in the future and restrict a PR to a main change.

### DIFF
--- a/pyworkflow/em/packages/relion/viewer.py
+++ b/pyworkflow/em/packages/relion/viewer.py
@@ -649,7 +649,6 @@ Examples:
 #===============================================================================
     def _showFSC(self, paramName=None):
         #self._iterations = self._getListFromRangeString(self.iterSelection.get())
-        print("_showFSC_self._iterations",self._iterations)
         threshold = self.resolutionThresholdFSC.get()
         prefixes = self._getPrefixes()        
         nrefs = len(self._refsList)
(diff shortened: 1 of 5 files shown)


pconesa:
Clearly, our bad. I'm reviewing this now. Just to describe a bit better the facts I'm adding a few lines here.

The main error was here: https://github.com/I2PC/scipion/pull/1716/files#diff-3c0c9301f1a3605767d56056ec241354L251

We detected this a few weeks ago and there is a pending PR. We had tested many, but not all visualization cases, including COORDINATES. And it worked (in our case because we had the memory parameter in our config and therefore the bug at https://github.com/I2PC/scipion/pull/1716/files#diff-3c0c9301f1a3605767d56056ec241354L251 wasn't happening.

Still, I agree we missed the Xmipp picker case.

@delarosatrevin, I think the 

>  coordinates visualization and other cases

would have been fixed with the above-mentioned PR. Otherwise....I fail to see the other cases.

delarosatrevin:
There were also important bugs...
here: https://github.com/I2PC/scipion/pull/1724/files#diff-62ff050c84a6b85a36c9b69c91c022cdR72
and
here: https://github.com/I2PC/scipion/pull/1724/files#diff-ed259d5a517348d15060c999c451b8dbR331

pconesa:
yes, those were the picking ones, I agreed.

PR #1727, opened 2018-08-17
Fix error when deleting a project in manager.py

### LINKED ISSUE #1700: Error project window
I have detected a bug when we create a new project and we delete it. After delete the project an error appears when we try to open another project created.
screenshot from 2018-07-23 12-51-17

pconesa:
On behalf of @fonsecareyna82.

Closes #1700

### DIFF
--- a/pyworkflow/manager.py
+++ b/pyworkflow/manager.py
@@ -139,6 +139,7 @@ class Manager(object):
 
     def deleteProject(self, projectName):
         pwutils.path.cleanPath(self.getProjectPath(projectName))
+        os.chdir(self.PROJECTS)
 
     def renameProject(self, oldName, newName):
         os.rename(self.getProjectPath(oldName), self.getProjectPath(newName))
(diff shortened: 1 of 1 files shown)


delarosatrevin:
Thanks @fonsecareyna82 and @pconesa for dealing with this issue. 

The main problem here was that Project.create function change the current working directory to the new project path. This break the deletion after from the Projects GUI. 

Instead of the proposed solution (changing the cwd to the PROJECTS dir, that might no be the original directory), I made a commit with the following change: 
* store cwd
* create project
* restore cwd

I think this is more clean since does not assume any location and just restore the cwd.

pconesa:
@delarosatrevin , @fonsecareyna82 this is causing tests to fail. @ldelcano is doing some test now but seems to break the tests.

We can undo this or review what the test are doing?

delarosatrevin:
Yes, maybe these change is affecting tests (some or all?) that were assuming this side effect of moving to the project folder. I will propose to keep the change and review the failing cases. I would prefer to keep the new behaviour of not changing the current location when creating a project.

ldelcano:
Yes, José Miguel, it is affecting all tests. I do not know if also when creating workflows from a json.

PR #1735, opened 2018-08-19
add dose to html report

### LINKED ISSUE #1460: HTML report: add dose params
I think it would be good to see in html report (Acquisition params) the dose (per frame and initial if not 0.0), unless user imports mics and not movies

azazellochg:
closes #1460


pconesa:
Don't know why so many changes have appeared..umm

PR #1736, opened 2018-08-19
Do not divide by the gain in xmipp movie alignment protocols

pconesa:


pconesa:
This is to keep consistency with motioncor2 and Scipion python code about gain. What I'm still not sure is if it is possible to find gain files that, to apply it, a division is required instead of a multiplication.

PR #1737, opened 2018-08-20
Convert gain

pconesa:
Convert gain and dark images if needed in a first step. For all movie alignment protocols. Test all combinations and step creation and dependency.

### DIFF
--- a/pyworkflow/em/protocol/protocol_movies.py
+++ b/pyworkflow/em/protocol/protocol_movies.py
@@ -105,7 +105,7 @@ class ProtProcessMovies(ProtPreprocessMicrographs):
 
         # Gain and Dark conversion step
         self.convertCIStep = []
-        convertStepId = self._insertFunctionStep('_convertCorrectionImages',
+        convertStepId = self._insertFunctionStep('_convertInputStep',
                                                   prerequisites=[])
         self.convertCIStep.append(convertStepId)
 
@@ -117,7 +117,7 @@ class ProtProcessMovies(ProtPreprocessMicrographs):
                                  prerequisites=finalSteps, wait=True)
 
     # STEP to convert correction images if apply
-    def _convertCorrectionImages(self):
+    def _convertInputStep(self):
 
         movs = self.inputMovies.get()
 
@@ -133,31 +133,35 @@ class ProtProcessMovies(ProtPreprocessMicrographs):
         """ Will convert a gain or dark file to a compatible one and return
         the final file name. Otherwise, will return same passed parameter"""
 
+        # Get final correction image file
+        finalName = self.getFinalCorrectionImagePath(correctionImage)
+
+        if not os.path.exists(finalName):
+            # Conversion never happened...
+            print('converting %s to %s' % (correctionImage, finalName))
+            ImageHandler().convert(correctionImage, finalName)
+
+        return os.path.abspath(finalName)
+
+    def getFinalCorrectionImagePath(self, correctionImage):
+        """ Returns the final path to the correction image (converted or not)
+        or and exception correctionImage does not exists"""
+
         if correctionImage is None:
             return
 
         elif not os.path.exists(correctionImage):
-            print (correctionImage)
-        else:
-
-            # Get final correction image file
-            finalName = self.getFinalCorrectionImagePath(correctionImage)
-
-            if not os.path.exists(finalName):
-                # Conversion never happened...
-                print('converting %s to %s' % (correctionImage, finalName))
-                ImageHandler().convert(correctionImage, finalName)
-
-            return os.path.abspath(finalName)
-
-    def getFinalCorrectionImagePath(self, correctionFile):
+            raise Exception("Correction image is set but not present in the "
+                            "file system. Either clean the value in the import "
+                            "protocol or copy the file to %s."
+                            % correctionImage)
 
         # Do we need to convert?
-        convertTo = self._getConvertExtension(correctionFile)
+        convertTo = self._getConvertExtension(correctionImage)
 
         if convertTo is not None:
             # Get the base name
-            fileName = basename(correctionFile)
+            fileName = basename(correctionImage)
 
             # Replace extension
             fileName = pwutils.replaceExt(fileName, convertTo)
@@ -166,7 +170,7 @@ class ProtProcessMovies(ProtPreprocessMicrographs):
             return self._getExtraPath(fileName)
 
         else:
-            return correctionFile
+            return correctionImage
 
 
     def _insertFinalSteps(self, deps):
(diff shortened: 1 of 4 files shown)


pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+
         # Conversion step is part of processMovieStep because of streaming.
         movieSteps = self._insertNewMoviesSteps(self.insertedDict,
                                                 self.inputMovies.get())
         finalSteps = self._insertFinalSteps(movieSteps)
         self._insertFunctionStep('createOutputStep',
                                  prerequisites=finalSteps, wait=True)
 
+    # STEP to convert correction images if apply
+    def _convertCorrectionImages(self):
```
delarosatrevin: I will propose to change the name by 'convertInputStep'.
Some considerations for that suggestion are:
* A convention for 'steps' functions is to add the suffix 'Step'
* 'convertInputStep' is a usual function in protocols and it will give us space for other conversions in the future if needed, not only this one related to gain and dark
* Step functions usually do not start with underscore

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+
+        # Convert gain
+        gain = movs.getGain()
+        movs.setGain(self.__convertCorrectionImage(gain))
+
+        # Convert dark
+        dark = movs.getDark()
+        movs.setDark(self.__convertCorrectionImage(dark))
+
+    def __convertCorrectionImage(self, correctionImage):
```
delarosatrevin: Here the input parameters is called correctionImage and in the next one correctionFile, I will suggest more consistency with the naming

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+        """ Will convert a gain or dark file to a compatible one and return
+        the final file name. Otherwise, will return same passed parameter"""
+
+        if correctionImage is None:
```
delarosatrevin: I will propose to handle the exception cases (as gain is None or there is not need to convert) in a single place/function, now there are here and in the next function.

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+            return
+
+        elif not os.path.exists(correctionImage):
```
delarosatrevin: It the gain is set and does not exist...the protocol can not just simply ignore it, I suggest we need to raise an Exception in order to notify the user that something is wrong: either remove the gain parameter or copy the file to the expected path.

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+            print (correctionImage)
```
delarosatrevin: Remove this (debug?) print

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+            finalName = self.getFinalCorrectionImagePath(correctionImage)
+
+            if not os.path.exists(finalName):
+                # Conversion never happened...
+                print('converting %s to %s' % (correctionImage, finalName))
+                ImageHandler().convert(correctionImage, finalName)
+
+            return os.path.abspath(finalName)
+
+    def getFinalCorrectionImagePath(self, correctionFile):
```
delarosatrevin: Why this method is not "private" (starting with __) as the previous one?
Could you consider a shorter name?

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -102,14 +102,73 @@ def _insertAllSteps(self):
...
+
+        # Do we need to convert?
+        convertTo = self._getConvertExtension(correctionFile)
+
+        if convertTo is not None:
```
delarosatrevin: Related comment before, try to handle the exceptional cases where no conversion is required in the same place.

pyworkflow/tests/em/protocols/test_protocols_motioncor2.py
```diff
@@ -136,3 +138,138 @@ def test_qbeta_motioncor2_sel(self):
         self._checkMicrographs(prot)
         self._checkAlignment(prot.outputMovies[1],
                              (2, 6), [0, 0, 0, 0])
+
+
+# Some utility functions to import movies that are used
+# in several tests.
+class TestMotioncor2GainFile(BaseTest):
```
delarosatrevin: Would be possible to add a test with DM4 file, that was the one that triggered this error?

pyworkflow/em/protocol/protocol_movies.py
pconesa: This method is used in a test to share the logic.

pyworkflow/tests/em/protocols/test_protocols_motioncor2.py
pconesa: Yep.

pconesa: I tried but we can not generate DM4 from xmipp....any other SW you know of that can generate dm4?

pconesa: ok  eman can. Forgot about it

pconesa: umm seems it can't write? http://blake.bcm.edu/emanwiki/EMAN2ImageFormats

azazellochg: Dm4 can be generated by dm only :)

pconesa: windows, right?

pconesa: I think I've addressed all the comments @delarosatrevin except testing DM4. But I've tested manually that case and worked.

pyworkflow/em/protocol/protocol_movies.py
```diff
@@ -201,7 +269,7 @@ def _insertMovieStep(self, movie):
         movieStepId = self._insertFunctionStep('processMovieStep',
                                                movieDict,
                                                movie.hasAlignment(),
-                                               prerequisites=[])
+                                               prerequisites=self.convertCIStep)
```
delarosatrevin: Has this been tested? I think it should be self._convertInputStep

pconesa: Yes it was tested, and it worked. To be honest, I'm not sure if this is the standard way but is done this way:
https://github.com/I2PC/scipion/pull/1737/files#diff-e003db186daef298b84aa05019e4f3ceR107

delarosatrevin: Ah, ok. I was confused about this value and the step function name. I will merge this PR then and we test everything together.

PR #1739, opened 2018-08-23
Adapt plugin installation script to new structure

pconesa:
Hi, this is just some minor changes to be able to install binaries through the plugins using the existing installation script.

I know @delarosatrevin you are not happy with this script, but at least now we should be able to test the installation of the binaries through the plugins. We might discuss later if we should refactor the entry point or even the related code that manages the plugins installation(information extraction).


delarosatrevin:
Pablo...I'm on vacations until the 3th of Sept. I will try to take a look
then. Anyway I agree that you should go ahead and test the mechanism and we
can refactor later

PR #1743, opened 2018-08-30
Adding tests to check the queue system

dmaluenda:
4 tests are included to check a certain queue system defined by the global constant [QUEUE_PARAMS](https://github.com/dmaluenda/scipion/blob/4da03680861edc08bad7d6ab6be4cb1e5114da19/pyworkflow/tests/em/workflows/test_parallel_gpu_queue.py#L44). In this case is set for the Slurm queue system configured in the I2PC machine called torvalds.

It contains 4 tests:
- **scipion test pyworkflow.tests.em.workflows.test_parallel_gpu_queue.Test_Queue_Small**: launching the relion2Dclassification with GPU and 4MPI using the queue system.
- **scipion test pyworkflow.tests.em.workflows.test_parallel_gpu_queue.Test_Queue_ALL**: launching the relion2Dclassification with all the convinations (with and without) between GPU, 4MPI, threads (8 diff. types) using the queue system.
- **scipion test pyworkflow.tests.em.workflows.test_parallel_gpu_queue.Test_noQueue_Small**: launching the relion2Dclassification with GPU and 4MPI **without** use the queue system.
- **scipion test pyworkflow.tests.em.workflows.test_parallel_gpu_queue.Test_noQueue_ALL**: launching the relion2Dclassification with all the convinations (with and without) between GPU, 4MPI, threads (8 diff. types) **without** use the queue system.

In the two first tests (queued), we check that the job is included in the queue and we wait until it get out of that queue, concluding that the job is done.

The two last tests (no queued) can be useful to check certain behavior when the user try to skip the queue system.

In addition, the [launchProtocol() in tests.py](https://github.com/dmaluenda/scipion/blob/4da03680861edc08bad7d6ab6be4cb1e5114da19/pyworkflow/tests/tests.py#L109) is modified where it check if the protocol is finished to take into account when the protocol is released to the queue and, obviously, is not finished yet.

This is being very useful for testing the new machine torvalds (slurm), however it can be adapted for any other queue system.


pyworkflow/tests/em/workflows/test_parallel_gpu_queue.py
```diff
@@ -0,0 +1,327 @@
...
+            queueList = queueRaw.split('\n')
+            for queueLine in queueList:
+                # we will assume that if protId and jobId
+                # is in the string, the task is there
+                if jobId in queueLine and str(protId) in queueLine:
+                    isThere = Yes
+                    break
+            return isThere
+
+        def wait_until(condition, timeout, *args, **kwargs):
```
pconesa: There is a "generic" wait function here: https://github.com/I2PC/scipion/blob/devel-pluginization/pyworkflow/tests/test_utils.py#L63
It would be nice to have one, compatible with your case too.

dmaluenda: Yes. That is included here is a little bit more complex, we can try to merge with the generic one, but it's not easy since one is returning True/False depending on the time consumption, and the other just return when the condition or the timeout is reached... Any suggestion?

PR #1745, opened 2018-08-31
Bug in ShowJ

vabrishami:
If the amount of memory was None, ShowJ did not work.

### DIFF
--- a/pyworkflow/em/showj.py
+++ b/pyworkflow/em/showj.py
@@ -248,7 +248,7 @@ def getJavaIJappArguments(memory, appName, appArgs):
     """ 
     if memory is None:
         memory = getJvmMaxMemory()
-        print "No memory size provided. Using default: " + memory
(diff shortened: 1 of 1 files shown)


delarosatrevin:
Thanks Vahid! I think this issue has been already addressed and released in
1.2.1

vabrishami:
Good! But I remember if a bug has been solved in the release, then it should be merged into devel? Correct me if I am wrong.

delarosatrevin:
That's true, but I think @pconesa has been merging 1.2.1 into devel.

pconesa:
Hi all, I was waiting for a couple of PR to be merged before sending release-1.2.1. turns out the fix is stuck there. https://github.com/I2PC/scipion/blob/release-1.2.1/pyworkflow/em/showj.py

pconesa:
I'm closing this since the fix will eventually make it to deve. Thanks @vabrishami

PR #1754, opened 2018-09-10
New defocus estimation with and without phase plate

JaviMota:
I have added a try except when we estimate the ctf to avoid program fails if some of the ctf has raise an error.
The flag enable ucs4 in python installation is to use tensorflow.


dmaluenda:
This pull request is related with [this in Xmipp plugin](https://github.com/I2PC/scipion-em-xmipp/pull/6/files) and [this in Xmipp software](https://github.com/I2PC/xmipp/pull/8/files).

pyworkflow/em/protocol/protocol_micrographs.py
```diff
@@ -588,8 +588,11 @@ def _updateOutputCTFSet(self, micList, streamMode):
 
 
         for micFn, micDir, mic in self._iterMicrographs(micList):
-            ctf = self._createCtfModel(mic)
-            outputCtf.append(ctf)
+            try:
+                ctf = self._createCtfModel(mic)
```
delarosatrevin: Can you double-check that this not already handled at other level? As far as I remember the movies or micrographs protocols do not fails if one item fails. On the other side, generally  I don't like to hide errors without a good reason. ;)

JaviMota: The reason is to avoid the ctf protocol stops when one ctf estimation fails. It raise a message from the binary xmipp_error... I dont know what happen with ctffind and gctf (Pablo is going to check this) but we can put a message inside Scipion with some extra information ;)

dmaluenda:
This is approved and the related PRs in the xmipp-plugin and in the xmipp-software are already merged. Thus, is no inconvenient, once the conflict file is resolved, I will merge this.

PR #1756, opened 2018-09-11
Form fixes and tweaks

### LINKED ISSUE #1732: help button not displayed in GUI form when there are two params in Line
Also, besides Help button missing, the whole form is distorted :)

screenshot_20180818_223532

pconesa:
* not editable params are flat now, and keep the grey color
* widgets are better aligned, labels entries and buttons.
* sticky params attribute changed to be more responsive to size, especially width.
* Line param bug is fixed (no showing help buttons).
* some widgets have names to improve debugging.

Closes #1732 

Default load:

default

Responsive to width:

responsive

Fixed bugs:
fixed

NOTE: when activating the expert mode...scrolls are both active and the horizontal shouldn't. This was better before, but a small resize of its width fixed this and all widgets are shown without the need to maximize the window. Which some cases, with a large amount of expert params, was impossible to show them.

### DIFF
--- a/pyworkflow/gui/form.py
+++ b/pyworkflow/gui/form.py
@@ -606,7 +606,7 @@ class SectionFrame(tk.Frame):
         canvasFrame = tk.Frame(self, name="sectioncontentframe")
         configureWeigths(canvasFrame)
         self.canvas = Canvas(canvasFrame, width=625, height=self.height,
-                             name="sectioncanvas", bg="purple")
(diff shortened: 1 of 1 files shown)


azazellochg:
Wow @pconesa , that's impressive! Thank you very much for looking into this!

delarosatrevin:
Great job @pconesa !!! These changes are localized in the form.py code... don't you want to include them in release-1.2.1?

PR #1763, opened 2018-09-13
Bug fixing in xmipp's viewer.py: 'str' + 'int'

dmaluenda:
DataView.setMemory( ) and showj.getJvmMaxMemory() works with strings, however we want to increase the available memory to show movies in 1Gb.

I don't know if there is a cleaner way bu this fix the bug.


pconesa:
memory should be int...ummm

delarosatrevin:
I also found this bug during the EMBO course last week, but I didn't had
the time to report it at the moment.

pconesa:
Just change it to avoid that double casting. Rest is fine. I've tested with and without the variable in the config and launched Viewers (Movies, Particles, Coordinates, Pickings and picking wizards) for all cases.

PR #1764, opened 2018-09-13
fix compillation errors with java 7

pconesa:


pconesa:
Java 7 requires variables used in threads, declared outside to be final.

PR #1766, opened 2018-09-17
Instance of a class insted of class itself in getattr

vabrishami:
When I tried to add my plugin, this error raised:

unbound method validateInstallation() must be called with Plugin instance as first argument (got nothing instead)

So I did this pull request to solve this bug.

### DIFF
--- a/pyworkflow/protocol/protocol.py
+++ b/pyworkflow/protocol/protocol.py
@@ -1600,7 +1600,7 @@ class Protocol(Step):
         """
         try:
 
-            validateFunc = getattr(cls.getClassPackage().Plugin,
+            validateFunc = getattr(cls.getClassPackage().Plugin(),
                                'validateInstallation', None)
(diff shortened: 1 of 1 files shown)


pyworkflow/protocol/protocol.py
```diff
@@ -1600,7 +1600,7 @@ def validateInstallation(cls):
         """
         try:
 
-            validateFunc = getattr(cls.getClassPackage().Plugin,
+            validateFunc = getattr(cls.getClassPackage().Plugin(),
```
delarosatrevin: @vabrishami I think this is not necessary, the method should be a "classmethod" in Python (like static in C++), so, it should be a method of the class, not an instance of the class.

vabrishami: Yes, Miguel, you are right. Thanks!

PR #1774, opened 2018-09-19
Co installtensorflow

cossorzano:


install/script.py
```diff
@@ -210,7 +211,7 @@ def removeByPattern(pattern):
     'python',
     tar='Python-2.7.14.tgz',
     targets=[env.getLib('python2.7'), env.getBin('python')],
-    flags=['--enable-shared'],
+    flags=['--enable-shared','--enable-unicode=ucs4'],
```
azazellochg: I would be cautious about this, since matplotlib and some other libs have to be recompiled after this change. Unless we put the same flag for them as well into install script

dmaluenda:
The installation of tensorflow can be done in the defineBinaries in the Xmipp plugin. But the '--enable-unicode=ucs4' flag for the python installation should be done by Scipion. The only way that I see to avoid changing the Scipion installation is to use an xmipp's own python, but it is not pretty worth... I guess...

pconesa:
let's just compile python compatible with Tensor flow. And install tensorflow from xmipp.

delarosatrevin:
Ok, just double-check that it does not break the compatibility with already existing packages.

joshua4289:
Hi Everyone 
Is this for the Warp package or a new protocol in xmipp ? 

Sincerely
Joshua Lobo

pconesa:
Hi @joshua4289 , this is to allow the installation of tensorflow in scipion's python. New developments of xmipp are using tensorflow and form scipion we have to allow this to hapen.

dmaluenda:
- I will take the code from the istall/script.py to the defineBinaries() in the [Xmipp plugin](https://github.com/I2PC/scipion-em-xmipp/issues/12).
 - The flag for the python installation is included in the [New defocus estimation... PR](https://github.com/I2PC/scipion/pull/1754).
 - The 'extra g' in the viewer-related code is already fixed in devel-pluginization.

Thus, I will close this pull request.

PR #1775, opened 2018-09-20
close #1772

### LINKED ISSUE #1772: import relion particles does not read alignments
**Describe the bug**
Import from star file when it does not end with *data.star works, but no alignment info is loaded. I do have some star files created by Join Star job in relion. They are not coming from 3d ref or cls.

**To Reproduce**
1. Import any star files from relion (e.g., 3d ref) that does not end with *data.star
2. Alignment=None

[Here](https://github.com/I2PC/scipion/blob/release-1.2.1/pyworkflow/em/packages/relion/dataimport.py#L182) is the bug.

azazellochg:
I have checked import with/without data extension. Works for me. Import tests are also ok.

### DIFF
--- a/pyworkflow/em/packages/relion/dataimport.py
+++ b/pyworkflow/em/packages/relion/dataimport.py
@@ -202,10 +202,19 @@ class RelionImport():
             else:
                 self._classesFunc = None
         else:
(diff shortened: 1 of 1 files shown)


pconesa:
remember to redo this in the plugin!

PR #1783, opened 2018-09-27
Bug fixes and improvements after pluginization

dmaluenda:
- The first and most relevant change is the _pyworkflow/em/metadata/constants.py_ : 
   - We replace all `SOME_MDL = xmippLib.SOME_MDL` to a single line `from xmippLib import *`
   - This is done in order to be able to manage different xmipp-bindings versions in a dynamic way.
   - I tested it and it works fine.
   - **Please review this**

 - Improving the PluginNotFound warning to include where is called.
 - Simplify the deprecated xmipp ghost warning.
 - Renaming the volume Homogenizer to Enrich in a workflow test.
 - Renaming the ctfSelection to ctfConsensus in a workflow test.
 - Fixing imports in workflow tests.


pyworkflow/utils/utils.py
```diff
@@ -729,13 +729,20 @@ def getEnvVariable(variableName, default=None, exceptionMsg=None):
 
 
 def pluginNotFound(plugName, errorMsg='', doRaise=False):
```
pconesa: What happens with the doRaise?
Just now we found it useful to use doRaise in the "mixed test" when we start importing plugins and they are not installed.

dmaluenda: `doRaise = True`  should raise an error telling the same than the warning in the case that `doRaise = False`. Have you test it??

dmaluenda: Ouch! Now I see!! It's delete it... My fault... I will review it!

Thanks @pconesa !

pconesa: ;-)

dmaluenda: Fixed! :)

PR #1787, opened 2018-10-02
Solving the way to list and filter the tests

fonsecareyna82:
Solving the way to list and filter the tests


pyworkflow/tests/em/data/test_data.py
```diff
@@ -13,7 +13,7 @@
 try:
     from xmipp3.convert import *
 except:
-    pluginNotFound('xmipp')
+    pluginNotFound('xmipp', noRaise=True)
```
delarosatrevin: noRaise or doRaise?

fonsecareyna82: doRaise.....sorry

pyworkflow/tests/em/workflows/test_workflow_high_throughput.py
```diff
@@ -29,13 +29,13 @@
 from test_workflow import TestWorkflow
 from pyworkflow.utils import importFromPlugin
 
-XmippProtOFAlignment = importFromPlugin('xmipp3.protocols', 'XmippProtOFAlignment')
+XmippProtOFAlignment = importFromPlugin('xmipp3.protocols', 'XmippProtOFAlignment', noRaise=True)
```
delarosatrevin: review many other cases of noRaise...if this is the actual name of the parameters...I will strong propose to be more consistent and keep only doRaise or noRaise. I will vote for the first to avoid negation in the name.

fonsecareyna82: The method parameter definition is doRaise.....I'm changing the other definitions

PR #1789, opened 2018-10-05
Devel pluginization fixes

fonsecareyna82:


pyworkflow/object.py
```diff
@@ -1162,8 +1162,13 @@ def update(self, item):
         self._getMapper().update(item)
                 
     def __str__(self):
-        return "%-20s (%d items)" % (self.getClassName(), self.getSize())
-    
+
+        return "%-20s (%d items%s)" % (self.getClassName(), self.getSize(),
+                                         self._appendStreamState())
+
+    def _appendStreamState(self):
```
delarosatrevin: I really needed a function for this? Why not just a variable inside the str function below. I will also say that is better to be more explict "o" does not mean anything and it is easy to forget, why not something like "open set" ?
```
streamState = "" if self.isStreamClosed() else ", open set"
return "%-20s (%d items%s)" % (self.getClassName(), self.getSize(),
￼                                                    streamState)

File: scipion [hunk 2]
```diff
@@ -41,9 +41,9 @@ except ImportError:
     from configparser import ConfigParser, ParsingError  # Python 3
 
 
-__version__ = 'v1.2.1'
-__nickname__ = 'Claudio'
-__releasedate__ = '2018-10-01'
+__version__ = 'v2.0'
+__nickname__ = 'Diocletian'
```
dmaluenda: Just for fun:

Diocletian separated and enlarged the empire's civil and military services and reorganized the empire's provincial divisions, establishing the largest and most bureaucratic government in the history of the empire. [[source](https://en.wikipedia.org/wiki/Diocletian)] Split and reorganize like pluginization does XD

Great!

config/templates/scipion.template
```diff
@@ -20,45 +20,13 @@ SCIPION_PLUGIN_REPO_URL = http://scipion.i2pc.es/getplugins/
...
-GCTF_HOME = %(EM_ROOT)s/gctf-1.06
-# replace '_cuX.Y_' to certain installed cuda
-GCTF = Gctf-v1.06_sm_20_cu8.0_x86_64
-# GCTF_CUDA_LIB = None # Fill to override scipion CUDA_LIB if different
-IMAGIC_HOME = %(EM_ROOT)s/imagic
-LOCALREC_HOME = %(EM_ROOT)s/localrec-1.2.0
-LOCALREC_RELION_HOME = %(EM_ROOT)s/relion-1.4
-LOCSCALE_HOME = %(EM_ROOT)s/locscale-0.1
+# Define plugin variables here, e.g:
+# EMAN2DIR = %(EM_ROOT)s/eman-2.12
```
dmaluenda: Why eman is the only one?

pyworkflow/object.py
```diff
@@ -1162,8 +1162,13 @@ def update(self, item):
...
+        return "" if self.isStreamClosed() else ", o"
```
dmaluenda: Great! But I prefer somthing like `SetOfMicrographs (234 items, 1024x1024, 3.54A/px)*` , i.e. an asterisk at the end (out of the parenthesis)

However, I think that is very useful whatever it is indicated

delarosatrevin: With * with should include a notation then ;)
But I also agree that we should try to keep names short. Also this will be visualized in the object selector.

config/templates/scipion.template
delarosatrevin: Yes, I think it should be removed as well.

scipion
delarosatrevin: Yes, I was curious and read the Wikipedia...really fits with the pluginization purposes...hehehehe.

josuegbl: I agree with the name, but I remember that we agreed to put it in SPANISH!!!!!!

delarosatrevin: @josuegbl Do you even know how 'Diocletian' is in Spanish?

josuegbl: as far as I know, the name in spanish is Diocleciano.... 5 seconds googling....

delarosatrevin: I see...anyway...I would say that you could spend better your energy and time (5 seconds lost googling). Have you read the pluginization docs? Any suggestion about anything that worries you? ...

josuegbl: I do not like the tone that is taking this conversation ... anyway, knowledge is never wasted time. If I had something to add, I would have done it already. Last comment, I have thing more important to do.

delarosatrevin: By the way, maybe a English name will reflect better the more international community of the project...in any case, maybe we could vote again for it. ;)

PR #1796, opened 2018-10-14
move logo to plugins, add more funcs to inspect-plugins script

azazellochg:
1. Most logos have been moved from pw/resources to corresponding plugins. There are still few left (some packages are not pluginised) yet
2. Add new arg (info) to inspect-plugins.py script. Now there is some overlap with pw_protocol_list.py but I think it looks better. To use it, execute `scipion run python scripts/inspect-plugins.py eman2 info`

Result:
screenshot_20181014_145936

I think this output can be improved/modified, so that we can auto-update https://github.com/I2PC/scipion/wiki/Integrated-Protocols page (it is outdated by a few years now..)


scripts/inspect-plugins.py
```diff
@@ -103,6 +122,75 @@ def getSubmodule(name, subname):
...
+        showBase = True if (n == 4 and sys.argv[3] == '--showBase') else False
+        subclasses = {}
+        emCategories = [('Imports', em.ProtImport),
+                        ('Micrographs', em.ProtMicrographs),
+                        ('Particles', em.ProtParticles),
+                        ('2D', em.Prot2D),
+                        ('3D', em.Prot3D)]
+
+        plugin = Domain.getPlugin(pluginName)
+        version = PluginInfo('scipion-em-%s' % pluginName).pipVersion
```
pconesa: This "info" part will only work for plugins that are "pypizised". Actually for the binaries I would use Plugin.defineBinaries() instead.

azazellochg: but this is plugin version here (which is only defined in pypi files), not binaries version..

delarosatrevin: I will vote for defining the version in "standard way" (I think not a standard, but a suggestion and common in many Python packages), i.e using `__version__` in the `__init__.py ` of the main package. I would prefer to have this information independently of PyPi and its infrastructure.

pconesa: If pip package metadata can take the version from there..sound good. But I think that is another issue.

delarosatrevin: Yes, that would be ideal...but not sure if possible.

PR #1818, opened 2018-11-04
ctf estimation in batch

azazellochg:
I've started to work on ctf in batch. For now I only focused on gctf, without looking at ctffind or xmipp protocols. 

This PR most likely breaks ctffind/xmipp ctf protocols, so I appreciate any help to do this.


azazellochg:
I will put back micName dir per mic / bathc of mics (in tmp folder only) since we agreed this is the best solution for parallel processing.

delarosatrevin:
Also related to this one:
https://github.com/I2PC/scipion-em-xmipp/pull/32

PR #1821, opened 2018-11-06
Jmrt pluginization

delarosatrevin:
* Simple function to allow execute targets from the Environment class
* Renamed inspect-plugins.py to plugins.py
* Use of argparse for better parameters handling
* Added some info/tools for dealing with plugin binaries


scripts/plugins.py
```diff
@@ -0,0 +1,239 @@
...
+    showBase = verbosity > 0
+    subclasses = {}
+    emCategories = [('Imports', em.ProtImport),
+                    ('Micrographs', em.ProtMicrographs),
+                    ('Particles', em.ProtParticles),
+                    ('2D', em.Prot2D),
+                    ('3D', em.Prot3D)]
+
+    plugin = Domain.getPlugin(pluginName)
+    version = PluginInfo('scipion-em-%s' % pluginName).pipVersion
```
yaizar: This relies heavily on the name convention, a bit of a dangerous assumption in my opinion. We should avoid this practice of finding things by adding the "scipion-em" at the beginning I think, seems a bit hacky.

scripts/plugins.py
```diff
@@ -0,0 +1,239 @@
...
+                cat = 'Base prot'
+
+        # skip Base protocols if not requested
+        if prots[prot].isBase() and not showBase:
+            continue
+        else:
+            print("%-35s %-35s %-10s %-s" % (prot, label, cat, desc))
+
+
+def installBinary(plugin, target):
```
yaizar: I don't see the need of having an install binary in two places. I'm modifying the install plugin script so it uses the Domain to install binaries i.e. won't use the distribution name for that. This is redundant and I think this is not the correct place to have an installation tool.

scripts/plugins.py
```diff
@@ -0,0 +1,239 @@
...
+    env = Environment()
```
yaizar: This won't work if the addPackage has defined the parameter deps. Please take into account that we have programmed the installb script to work with all plugins, if you need a parallel way of installing plugin binaries that works for you I think it should not live here. Why is `scipion installb someBin`  worse than `scipion run scripts/plugins.py somePlugin --install someBin`? Why do we need both?

scripts/plugins.py
yaizar: One way to deal with this would be to add the possibility of accessing distributions within the Domain.

delarosatrevin: @yaizar The version obtained here from PluginInfo class is only used in a print statement below. Do you seriously consider this dangerous? This was there in the script before, not introduced in this PR. In a more general scope, I think the version should be taken from the Python package and not from the distribution metainformation.

scripts/plugins.py
delarosatrevin: After two days of trying to install/update the pluginazed version, I strongly need to test installation of new version of binaries...which was not supported when the plugins were available in the Python path. Once it is working...it is not longer needed. 
I

delarosatrevin: I think the Environment internally take care of targets dependencies, so I think it will work if the package binary has dependencies.

azazellochg:
what happened? should I still review it?

delarosatrevin:
Thanks @azazellochg! I just closed the PR, revision is not longer need.

scripts/plugins.py
yaizar: It would have taken you way less than two days to pipify the plugin but I'll leave that discussion be x) won't matter with changes I'll commit now

delarosatrevin: Yaiza, I don't know how long it would take me to pipify some plugin, or update the package or whatever. I just consider that is unacceptable to force people to pipify while developing and testing a plugin.

pconesa: Sad! :-(

PR #1822, opened 2018-11-06
Fix incorrect named argument to irecv.

akesandgren:
Irecv uses "source" not "dest".

Do you prefer to get PR's against master or devel branch?


delarosatrevin:
Thanks @akesandgren for reporting this. I think it is fine to master...if the issues are critical we might consider to release a bug-fix version....devel should be used for bigger changes.

PR #1823, opened 2018-11-07
ITMAX and IPMAX arguments to frealign are integers not floats.

akesandgren:
Without this change frealign crashes with
00370:   ./block001.sh
00371:   At line 34 of file card2.f
00372:   Fortran runtime error: Bad integer for item 9 in list input


akesandgren:
Note, that pre-existing project.sqlite databases that uses frealign will contain incorrect data and need to be manually updated.
I have a small script that does this.

azazellochg:
@pconesa I think it was a bad idea to modify protocol param type here.  I had to trace this error today, almost half a year after this PR. Breaking backward compatibility with old projects might be acceptable for a major release, but for minor ones it should have been done by fixing conversion step.

akesandgren:
@azazellochg First of all it's only 3 months since the PR, and without it the code is completely broken and thus useless. And the latest release is still older than the fix, so there has not been a new release with this fix.

PR #1835, opened 2018-11-21
Data export tweaks

yaizar:
PR to be able to recover relative paths from a workflow json for reproducibility purposes, done taking into account scipion-em-empiar


pconesa:
@delarosatrevin , this was work done by @yaizar long time ago for the EMPIAR deposition. The main change is that a protocol has the change to customize how to import/export the workflow. And the import images is doing it to make the paths relative and export the data too.

As an extra....the is a script to stop a project  , requested by eBic.

delarosatrevin:
@pconesa, Changes here does not seems very harmful, so I don't want to start another debate and I'm fine merging them.

For the future, I think that the import/export logic should not go into the protocols code. Protocols code is mainly concerned to processing data and generating outputs as defined in the data model. I think that import/exports should be in a separated layer on top of that infrastructure. Image that I want to export now a project to hdf5...or any other convention/format (not json), should I then modify the protocols? or if in another context I don't want the path to be relative, but mapped differently? 

Anyway, we can merge these changes as they are now. Just make sure that related test are passing.

pconesa:
I see your point...but doing it "right" would have more impact on the code and require more time. Thank you.

delarosatrevin:
Yes, I understand...so let's go ahead and merge this PR.

PR #1839, opened 2018-11-29
Extract coordinates in streaming

dmaluenda:
Allowing the extract coordinates protocol to be able to operate in streaming.


dmaluenda:
Yes @pconesa, it would be nice to have a test for this protocol. I fact, I thought that test already existed (for the non-streaming version), but it didn't. Now, I don't have enough time to do it, but some weeks later I will do it.

pconesa:
It stays a long time Launched. We need to revisit this to improve performance.

delarosatrevin:
> Yes @pconesa, it would be nice to have a test for this protocol. In fact, I thought that test already existed (for the non-streaming version), but it didn't. Now, I don't have enough time to do it, but some weeks later I will do it.

@dmaluenda Several weeks has passed since @pconesa suggested to add a test, has you had some time for it?

delarosatrevin:
Moreover, I think the performance is really bad and this is unacceptable to compromise the non-streaming case becase of the streaming. 

For example, I had a set of ~75K particles (not really big for these days standards) and the normal extract-coordinates took less than 2min...in this branch, I have executed the same jobs (same parameters) and it has been several minutes in "launched" without even start computing. I don't want to even try with 200k, 500k or 1million particles.

This protocol for extracting the coordinates is more to go to the next level when you have clean up your particles and want to push resolution. In my opinion, it is no very useful in streaming mode. But this is just my opinion, so, I'm fine with the streaming implementation. But as I said, it can not make the more common case to work much slower.

azazellochg:
What are the use cases for this protocol in streaming?

dmaluenda:
Yes @delarosatrevin , the main reason why this PR is not accepted yet is because its performance issue. The test was missing also with the non-streaming version of the protocol, but as you said, the problem in the performance is intolerable... I have to work more in that... I will try to fix this in 2 weeks, but I promise nothing ;-p

@azazellochg the extract coordinates in streaming may be useful in workflows where you want to re-extract particles from micrographs that have passed through different preprocess protocols than the original ones or when the direct way, extracting particles from other micSet, is not possible.

In fact, the subset protocol is not streamified and it would be nice in order to increase the possible complexity in the streaming data procesing.

dmaluenda:
I've just fixed the performance issue. ;-)

delarosatrevin:
Thanks @dmaluenda for improving the performance. 

Still it is worse than before (approx. timing). 
Particles: 100K old: 30sec -> new: 1min
Particles: 200K old: 50sec -> new: 2min

So it is now twice slower. It is not that bad, but I'm not sure if we should make the more used case (e.g 98% of cases) slower just because the streaming one (~2%). I still do not understand the lost of performance since the work to be done should be the same. So, I would propose the following:

1) Treat the no-streaming as special case (validate at the insertSteps) and fallback to the old implementation, so there will not be performance penalty when non in streaming.
2) Create a new protocol specific for the streaming case.

dmaluenda:
I will follow the first. Still, I don't understand where is losing that time....

delarosatrevin:
Great David! I don't understand neither why is the difference. I haven't had the time to go really deep in the code details.

dmaluenda:
Well, the splitting is done. Indeed, it's taking a half of the time in the nonStreaming mode... The reason is because, firstly, a loop over particles is done to be sure that the corresponding micrographs is available (maybe it's not in the input, yet), then a second loop is done in the extractStep..

delarosatrevin:
Great! Merged and branch deleted! Good to know the reason of why was slower.

PR #1840, opened 2018-11-29
Master to devel

pconesa:
Updating devel with master. Grigorief plugin has already this change.


pconesa:
Can anyone check this PR, it's just, I hope the last update of devel from master.

PR #1841, opened 2018-11-30
Merging devel-plug-fixes 

dmaluenda:
- Including the 'open set' label to SetOfCoordinates at the end of the set info in the _str_ casting (Summary outputs tab)
 - Adding the original protocol as prerequisite in the created protocols by 2d streamer
 - A join of SetOfClasses doesn't make sense (at least it fails so far, and the essay fix doesn't work), thus we assert if the input are SetOfClasses and if so raise a validation error.
 - Using the [pluginNotFound()](https://github.com/I2PC/scipion/blob/d3d576d1c4e6d0ac8c7d3c9dcba7d81b694402ac/pyworkflow/utils/utils.py#L731) function when an import fails in _pyworkflow/plugin.py_ in order to give more hints why/where is failing.


pconesa:
@delarosatrevin thanks!

PR #1846, opened 2018-12-11
install mpi4py 3.0.3

pconesa:
previous mpi python version didn't work for modern openMPI versions like 3.0. This has fixed my errors.

I've trested mpi with scipion test xmipp3.tests.test_protocols_xmipp_mics.TestXmippExtractParticles
But none of the protocols inside used mpi. So I manually run one with mpi.

use source, instead of dest for irecv
fix bug in test.py --> useQueue()

### DIFF
--- a/pyworkflow/tests/tests.py
+++ b/pyworkflow/tests/tests.py
@@ -106,7 +106,7 @@ class BaseTest(unittest.TestCase):
             print "    FAILED with error: %s\n" % prot.getErrorMessage()
             raise Exception("ERROR launching protocol.")
 
-        if not prot.isFinished() and not prot.useQueue:  # when queued is not finished yet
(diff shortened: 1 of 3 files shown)


pyworkflow/tests/tests.py
```diff
@@ -106,7 +106,7 @@ def launchProtocol(cls, prot):
             print "    FAILED with error: %s\n" % prot.getErrorMessage()
             raise Exception("ERROR launching protocol.")
 
-        if not prot.isFinished() and not prot.useQueue:  # when queued is not finished yet
+        if not prot.isFinished() and not prot.useQueue():  # when queued is not finished yet
```
dmaluenda: Auch! Sorry!

PR #1856, opened 2018-12-22
Add indexes and pragma instruction for database

rmarabini:
The **aim** of this commit is to **speed up searchs in metadatas.** Jaime Martin Benito reported that extracting particles form a large dataset (6000 micrographs and 150 coordinates per micrograph) spent several hours converting the coordinate fle (in sqlite) to xmipp (pos) coordinate files.

The more consuming code was the loop

```
        for micKey, mic in micDict.iteritems():
            if counter % 50 == 0:
                b = datetime.now()
                print b-a, 'reading coordinates for mic number', "%06d" % counter
                sys.stdout.flush()  # force buffer to print
            counter += 1

            micId = mic.getObjId()
            coordList = []
            self.debug("Loading coords for mic: %s (%s)" % (micId, micKey))
            for coord in coordSet.iterItems(where='_micId=%s' % micId):

```
In order to reduce the serching time we explore the posibility od adding pragma to the database and indexes.

=========================================================
**1st step)**

Follows a description of the preliminary research: first I measure the number of rows per second that sqlite may read/write under different conditions. With that aim I used the files 

- [select.txt](https://github.com/I2PC/scipion/files/2705332/select.txt)
- [bench.txt](https://github.com/I2PC/scipion/files/2705333/bench.txt)

_Bench_ inserts row under different conditions while _select_ search for rows. Bench shows that choosing properly the value of pragmas:

- 'synchronous': 'OFF'
- 'journal_mode': 'OFF'
- 'temp_store': 'MEMORY'
- 'cache_size': '5000'

Conclusions:

- you can improve insert speed by a factor of 20 (for 1000000 rows).  
- On the other hand,  _Select_  (for 1000000 rows) can be speed up by a factor of 100 using indexes
- Inserting when two indexes are defined is about 20% slower

========================================================================
**2nd Step**
 
Then using the code provided in test_data.py: TestSetOfCoordinates I compared insertion speed under different pragma values. No difference in insertion time was measured. After that I study how scipion saves data do sqlite databases and mimic it in the becnh.py file (the more important feature is that scipion does not commit the insertions until the setofObjects was been fully created). becnh.py suggested that choosing the pragma properlly a speed improvement of 2 should be obtained. 

Conclusion is: 

- In insertion mode the time python spends handling the python objects is greater than the time needed to save the data to the databse, 
- Optimizing the database access parameters for insertion will not improve the insertion rate 
- create indexes will not penalize insertion time since the database spends more of the time waiting for the next insertion.
============================================================================
**3rd step** 
The situation is different for select. Searching is much faster with indexes. In fact, the search that required several hours in "protocol_particles.py" now it done in 1 minute.
-----------------------------------------
Conclusion: Insertion is not improved by wise use of pragmas but search must be,

IMPORTANT:  "protocol_particles.py"  is greatelly improved by the use of indexes but...  "protocol_particles.py" is poorly written. If we rewrite the code as

```
            for coord in coordSet.iterItems(orderBy='_micId',
                                                   direction='ASC'):
                micId = coord.getMicId()
                if micId != lastMicId:
                    lastMicId = micId
                    .-.-
                .-.-

```
then the use of indexes only improve the search time in 15%.

In summary, it is a good idea to active indexes by default since they do not slow insertion and
may improve a lot the code. If the code has been carefully written this improvement may be small but for poorly written code the improvement may be dramatic

TEST: in file test data search for test_mapper tests

### DIFF
--- a/config/templates/protocols.template
+++ b/config/templates/protocols.template
@@ -212,7 +212,6 @@ Protocols SPA = [
 				{"tag": "protocol", "value": "ProtRelionSubtract",  "text": "default"},
 				{"tag": "protocol", "value": "XmippProtConvertToPseudoAtoms", "text": "default"},
 				{"tag": "protocol", "value": "AtsasProtConvertPdbToSAXS", "text": "default"},
-                {"tag": "protocol", "value": "XmippProtCombinePdb", "text": "default"}
 			]}
 		]},
 		{"tag": "protocol_group", "text": "Reconstruct", "openItem": "False", "children": [
--- a/config/templates/protocols.template
+++ b/config/templates/protocols.template
@@ -211,7 +211,7 @@ Protocols SPA = [
                 {"tag": "protocol", "value": "XmippProtSubtractProjection", "text": "default"},
 				{"tag": "protocol", "value": "ProtRelionSubtract",  "text": "default"},
 				{"tag": "protocol", "value": "XmippProtConvertToPseudoAtoms", "text": "default"},
-				{"tag": "protocol", "value": "AtsasProtConvertPdbToSAXS", "text": "default"},
+				{"tag": "protocol", "value": "AtsasProtConvertPdbToSAXS", "text": "default"}
 			]}
 		]},
 		{"tag": "protocol_group", "text": "Reconstruct", "openItem": "False", "children": [
--- a/pyworkflow/em/protocol/protocol.py
+++ b/pyworkflow/em/protocol/protocol.py
@@ -53,7 +53,7 @@ class EMProtocol(Protocol):
     def __init__(self, **kwargs):
         Protocol.__init__(self, **kwargs)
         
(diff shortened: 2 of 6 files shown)


pconesa:
Many thanks @rmarabini, very impressive and detailed explanation of the issue. I'll  review this.

config/templates/protocols.template
```diff
@@ -212,7 +212,6 @@ Protocols SPA = [
 				{"tag": "protocol", "value": "ProtRelionSubtract",  "text": "default"},
 				{"tag": "protocol", "value": "XmippProtConvertToPseudoAtoms", "text": "default"},
 				{"tag": "protocol", "value": "AtsasProtConvertPdbToSAXS", "text": "default"},
```
pconesa: Forgot to remove last comma, please!

PR #1865, opened 2019-01-02
Dynamic protocols tree

fonsecareyna82:
The project protocols treeView is created from the protocols.conf file that must be added in each of the plugins (scipion-em-plugin/plugin/protocols.conf). 
This file is similar to the one located in scipion/config/protocols.conf


pyworkflow/config.py
```diff
@@ -150,52 +156,78 @@ def addToTree(menu, item, checkFunction=None):
...
-            # If it is a protocol
-            if item["tag"] == "protocol":
-                # Get the class name and then if it is disabled
-                protClassName = item["value"]
-                protClass = em.Domain.getProtocols().get(protClassName)
-                if protClass is None:
-                    return False
-                else:
-                    return not protClass.isDisabled()
+    pluginRepo = PluginRepository()
```
pconesa: Can't we use Domain methods here....

delarosatrevin: Yes, I agree that Domain methods should be better and using the location from the Plugin class, it can even have a method getProtocolsConf (or similar) that returns the protocols.conf file of this plugin...if not, we could just assume it will be named like that in the plugins folder.

PR #1866, opened 2019-01-02
dynamic_protocols_tree

fonsecareyna82:
Creating the protocol_tree_view from the plugins procols.conf file

### DIFF
--- a/pyworkflow/project/config.py
+++ b/pyworkflow/project/config.py
@@ -539,23 +539,22 @@ class ProtocolTreeConfig:
         protocols[cls.ALL_PROTOCOLS] = allProtMenu
 
     @classmethod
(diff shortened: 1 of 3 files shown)


pyworkflow/project/config.py
```diff
@@ -34,6 +35,7 @@
 import pyworkflow as pw
 import pyworkflow.object as pwobj
 from pyworkflow.mapper import SqliteMapper
+from pyworkflow.install.plugin_funcs import PluginRepository, PluginInfo
```
delarosatrevin: I guess this import is not longer needed, right? We should avoid unnecessary import from pyworkflow.install, since this code should run in many versions of Python.

PR #1867, opened 2019-01-03
Html report tweaks

pconesa:
Phase shift added
Performance improved (only loading page images and not all)
New chart: CFT Time Series
image


dmaluenda:
Great!! I prefer lines connecting points but as you prefer.

On the other hand, why phase shift axis is from -240º to 240º? If it goes from -90º to 270º (or any other ranging the whole 360º) then important values [0, 90º and 180º] will have the grid line instead of +/-120º. I think this would visually help the user.

pconesa:
@dmaluenda  thanks for the feedback, I've tried lines, and points (on Y) will become very close to each other, lines will be a mess. I think that is why others do not use lines.
 I left hightchart to decide based on values, but if we agree on a concrete scale for the y axis I'll change it. TBH, I do not know what are the Phaseshift limits.

azazellochg:
@pconesa I'll have a look this week-end.

pconesa:
@azazellochg only if you can, I just added you because your feedback in the past has been very useful, but don't want to make you work on the weekend.

PR #1868, opened 2019-01-03
Devel to pluginization

pconesa:
Last changes from devel (and master)...after this....pluginization should be able to go back to devel.


pconesa:
@delarosatrevin , do you mind to check this. This should be the last thing coming from devel.

PR #1892, opened 2019-01-10
Subset bug: missing properties in movies subset

### LINKED ISSUE #1852: gain reference information for movies is lost upon creating movie subset
Hi,

I noticed that when you create a movie subset the gain reference associated with the movies is lost and you need to specify it as an additional parameter in motioncorr2.

pconesa:
Closes #1852


dmaluenda:
There is another bug in the Subset protocol, if you want to use this PR to also fix it.. ;-p

The protocol fails when you try to make a subset of coordinates because the createSetOfCoordinates function takes one argument (a SetOfMicrographs object) whereas it take no argument for the rest of the sets. So, the following lines fail.

https://github.com/I2PC/scipion/blob/cf2f9d3a33f094eef8f7b8cd1375ac07fcfdb7ec/pyworkflow/em/protocol/protocol_sets.py#L427-L430

I think that taking that as an exception if the inputClassName is SetOfCoordinates and passing then the micSet of the inputCoords to the createSetOfCoordinates, it's enough to work fine.

pconesa:
Thank you, @delarosatrevin. I'm looking at your request.

pconesa:
and your too, @dmaluenda

pconesa:
I think I've addressed all the request here. Nevertheless. Subset of Coordinates wasn't the only one failing:
image

All the methods listed here with a different signature than (self, suffix) will fail (except Coordinates now fixed. I've tried the same approach with 2Dclasses but it didn't work since is not just creating the set, but something else has to be fixed. I do not have time to fix all, so:
1.- are we happy with this small improvement, although still work to do
2.- Should we limit the param classes to those that match the signature + coordinates
3.- Something else?

pconesa:
Basically, 2dclass, 3dClass and tiltpairs.

dmaluenda:
I think that the subset for setOfClasses doesn't make sense, thus I would limit this class either in the addParams or in the validation.

I'm not sure for the tiltPairs, I think that it can make sense, however maybe it's not easy to fix. Therefore, I will say that just limiting it at this moment, as well. And in a future we can manage this.

pconesa:
I've added a validation to not allow sets that will fail : 2dclasses, 3d classes and tiltpair.
@dmaluenda , @delarosatrevin .

delarosatrevin:
Hi @pconesa and @dmaluenda , 
The following test is not failing explicitly, but a validation is failing inside the protocol:
scipion test pyworkflow.tests.em.protocols.test_protocols_sets.TestSets.testJoinValidation

```
...
Exception: Protocol.run: Validation errors:
There are different dimensions among the input sets: (8256, 9632, 1) and (9216, 9441, 1)
```

pconesa:
Yes, that is intentional. I wanted to test that the validation takes place and raises an error. It is the expected behaviour.

delarosatrevin:
Ok, great, then everything seems fine.

Issue #1185, opened 2017-08-25
IgnoreDuplicates not used?


cossorzano:
In pyworkflow/em/protocol/protocol_sets.py, ProtUnionSet there is a form field called ignoreDuplicates, but it seems not to be used in the implementation

pconesa:
This is old, but need to check the status.

pconesa:
I'm closing this issue. Feel free to reopen it if I'm wrong or missed something. It's working in devel.

Issue #1894, opened 2019-01-11
Ignore duplicates option on JoinSets protocol is not implemented


dmaluenda:
**Describe the bug**
 In the JoinSets protocol , there is an Advanced option to ignore duplicates but this parameter is not used in the protocol implementation. It does NOTHING!

**To Reproduce**
* In the GUI:
  * Classify a set of particles with two independent classifiers
  * Open the results and select some classes and make a setOfParticles by clicking on the '+Particles'
  * Do the same for the other classifier
  * Join the created sets with and without the 'ignore duplicates' options
  * See that the results are identical.
* In the CODE:
  * Go to the [JoinSets protocol code](https://github.com/I2PC/scipion/blob/8423ca7d43b199944020bec00c10de9c9269d47d/pyworkflow/em/protocol/protocol_sets.py#L110).
  * Look for (ctrl+f) 'ignoreDuplicates'
  * See that this parameter is not used.

**Desktop (please complete the following information):**
 - Version 2.0

PR #1895, opened 2019-01-11
Allowing "ignore duplicates" for the joinSets protocol

dmaluenda:
Now when "ignore duplicates" is selected, the protocol avoids to include duplicated items (it only includes the first item found).

In addition this option is always shown (not in the advanced mode) since it is clear to be used and quite usual.

Moreover, the label and the help for the renumbering option is improved.

This closses #1894


pconesa:
Closes #1185

dmaluenda:
mmm... Sure! I agree with that. I will include that validation.

pyworkflow/em/protocol/protocol_sets.py
```diff
@@ -236,11 +246,49 @@ def _validate(self):
...
+
+            # For pointer to a set
+            for setPointer in self.inputSets:
+
+                # Get the set:
+                inputSet = setPointer.get()
+
+                # If the set has the attribute
+                if not hasattr(inputSet, attr):
+                    break
```
dmaluenda: Why this break? Could it be for the case of setOfCoordinates, which hasn't getSampligRate() nor getDimensions()? 

In that case, the protocol also fails fails due to the line 136-138 (the `_create%(className)s`)... Then, two options are there either fixing the setOfCoordinates error as in [the subSet protocol](https://github.com/I2PC/scipion/pull/1892/commits/f31b33b3e510033e280be7506b07f49ce20746bb) or not passing the validation.

pconesa: The break is just in case the Set does not have the attribute, then the validation of that attribute is skipped. I didn't go through all combinations of attributes, I just thought is safe to have a list, and if the set does not have the attribute , move on and check the next attribute.

pconesa: what is the sert of corrdinates error?

dmaluenda: Great! But maybe a continue would be better. 

> The break is just in case the Set does not have the attribute, then the validation of that attribute is skipped. I didn't go through all combinations of attributes, I just thought is safe to have a list, and if the set does not have the attribute , move on and check the next attribute.

dmaluenda: The same as in the subSet protocol that you fixed [here](https://github.com/I2PC/scipion/pull/1892/commits/f31b33b3e510033e280be7506b07f49ce20746bb). When you try to do any set operation with SetOfCoordinates they fail becasue it's trying to launch _createSetOfCoordinates with wrong signature.

Well, we can skip this.

> what is that set of coordinates error?

pconesa: there are 2 loops:
1.- by attribute
      2.- by all inputset

           break is here inside 2

I wanted to break loop number 2, since it does not make sense to continue checking the rest of the sets.

I've just testes if the brake breaks the 2 loops and it doesn't

PR #1899, opened 2019-01-14
fix wrong columns 

azazellochg:


pconesa:
@azazellochg thank you....since what I did was working for me, but not for you.....my guess here is that we need a more robust way of getting the values instead of an integer (5,6 or 6,7) I guess you had the shifts column, whereas I didn't....I'll try to get CC values and resolution in a more robust way.

pconesa:
he, he never thought about that....we are taking the ctf creation time (checkNewOutput? ), and I guess they are created at the same time. Probably the best value to use would be movie file creation timestamp?.... Another option is not to use a timestamp but an index, but that will "hide" mic pauses (this was our thought to reject using micindex on the x axis).

azazellochg:
@pconesa one more thing, I have this undefined text showing up in both mic and drift thumbnails... any idea?
screenshot_20190115_140415

azazellochg:
I think using movie file creation time is the best option, this way it is more linked with the acquisition

pconesa:
@azazellochg , I've addressed 2 things here except using the movie ts, but I've created an issue to be addressed later. #1904

pconesa:
If you are ok, I'll will merge this.

azazellochg:
thanks! you can merge it!

PR #1902, opened 2019-01-15
heterogeneity branch in pluginized version

ajimoreno:
In this branch we created several protocols to manage heterogeneity and a "metaprotocol" able to run all these protocols with a scheduler to automate the process.


delarosatrevin:
@ajimoreno After the merge of the PR with the new way of specifying the protocols.conf for each plugin, you don't need to modify the global protocols.template anymore. You should modify the protocols.conf of xmipp plugin in this case and add these new protocols. I'm closing this PR.

PR #1905, opened 2019-01-22
Import changes

### LINKED ISSUE #1702: use regular expression in import to avoid importing certain files
I though this request was already an issue but I cannot find it.

Modify import movies so some directories can be skipped. Wildcards are not enought for this function but regular expressions will do the trick. 

A possible way to do this is to  get a list with all the files under a directory and then apply the regular expresion to the list.

I look to the import movie code and the relevnt methods seems to be in file base.py directory protocol_import.  See for example: getPattern. Unfortunately, I do not understand the code there. For example

        pattern = expandPattern(fullPattern.replace("$", ""))
        match = re.match('[^#]*(#+)[^#]*', pattern)

what are we doing in the second line and why we do need the first one?

### LINKED ISSUE #1898: Remove streaming socket
Remove the import via streaming socket that was added because of a facility where it's finally not being used.

### LINKED ISSUE #1903: Import images: repeated calls to _validateImages
It seems that in import protocols, we iterate over all the images that match the pattern a couple times before launching the import protocol in the function [_validateImages](https://github.com/I2PC/scipion/blob/master/pyworkflow/em/protocol/protocol_import/images.py#L333). Sometimes we've heard from users that imports take some time to start, so it could be because of this (we're validating the whole list of images twice). Maybe iterating the whole list for the validate is a bit excessive? maybe each image could be validated individually at the import time, not before, and when pressing "execute" we could validate only the form params? 

There's other validate instances where we only do it in the first one, like [this _validate](https://github.com/I2PC/scipion/blob/master/pyworkflow/em/protocol/protocol_import/micrographs.py#L230)

yaizar:
- Remove streaming socket, closes #1898 
- Adds blacklist to import movies and micrographs, closes #1702
- Call _validateImages only on first image, dont iterate all images before launching protocol, closes #1903


yaizar:
@pconesa changes have been addressed

pconesa:
Thanks @yaizar. I like how this has ended up. Nice.

PR #1909, opened 2019-01-29
Allow scalars to hold internal pointer to grab referenced value

delarosatrevin:


delarosatrevin:
Reopened here: https://github.com/I2PC/scipion/pull/1912

PR #1911, opened 2019-01-31
Resume all

fonsecareyna82:
* Functions to Restart or Continue the Project WorkFlow


pconesa:
it might be dirty but the case here is that if an input is a SetOfCoordinates, it is not enough to update the "owner" of the SoC, but we need to update also the protocol that is pointed by the SoC ( self._micrographsPointer = Pointer())
Maybe is something to add to the: project._updateProtocol ?

PR #1916, opened 2019-02-06
Support preferred viewers

### LINKED ISSUE #1850: Bsoft viewer is now the default volume viewer and chimera doest't open from the icon
**Describe the bug**
From viewers refactorization, the default viewer for volumes becomes the Bsoft viewer. Additionally, when the Scipion viewer is chosen by right-click on a volumeObject, chimera doesn't open by clicking at the icon at top-right.

**To Reproduce**
1. Select a protocol with a volume as main output (i.e. Reconstract significance) 
2. Click on Analyze Results icon
3. See the Bsoft viewer opened with the outputVolume  **<- First issue!**
4. Close the Bsoft viewer
5. Right-click on the outputVolume
6. Select "Open with DataViewer"
7. Click on chimera icon at top-right
8. See that nothing happen, even no error in terminal  **<- Second issue**

**Desktop:**
 Version 2.0 Diocletian

### LINKED ISSUE #1915: Spefific viewer is not launched if protocol inherits from a target of DataViewer.
**Describe the bug**
If a plugin define a specific viewer for a certain protocol (via target) and, at the same time, this protocol inherits from a protocolClass defined as a target for the DataViewer, that specific viewer is ignored when 'Analyze result' is clicked.

The reason is because `pyworkflow/em/__init__.py.findViewers()` (lines from 78 to 89) returns a list where the DataViewer is always first. And then, that first item is used to visualize the data.

**To Reproduce**
1. Open a project which contains a Xmipp**Auto**Picking.
2. Select the  XmippAutoPicking protocol and click 'Analyze result'.
3. See the 'State' column with all micrographs set as 'Manual'.
4. Close the project
-------------------
5. Edit `pyworkflow/em/__init__.py.findViewers()` (lines from 78 to 89) including the following lines before the return (**solution suggested till a more flexible and elegant solution is implemented**)
```
pyworkflow/em/__init__.py:

    .../...
    
89:    from pyworkflow.em.viewers import DataViewer
90:    if len(viewers)>1 and isinstance(viewers[0], DataViewer):
91:        viewers = reversed(viewers)
```
6. Repeat 1. and 2.
7. See the 'State' column with different states corresponding to the trainning process and Xmipp Staff.

**Expected behavior**
If a specific viewer is found, ignore the general one. (under my point of view)

**Additional**
In the case described above it is not a big deal, but we are finishing a protocol which is also inheriting from ProtocolAutoPicking base class and we have design a specific viewer **whose behavior is completely different from the general one** (in fact, the general one doesn't make sense...)

yaizar:
Supports adding preferred viewers to the user's scipion.conf with the variable VIEWERS e.g.:

```
VIEWERS={"Volume":["bsoft.viewers.BsoftViewer"],
         "SetOfVolumes": ["bsoft.viewers.bsoft_viewers.BsoftViewer"],
         "XmippParticlePickingAutomatic" : ["xmipp3.viewers.viewer.XmippViewer"]}
```
- closes #1915 
- closes #1850


pyworkflow/em/__init__.py
```diff
@@ -77,16 +76,24 @@ def findSubClasses(classDict, className):
 
 def findViewers(className, environment):
     """ Find the available viewers for this class. """
+    from pyworkflow.em.viewers.viewers_data import DataViewer
```
delarosatrevin: I don't think this import is necessary here...since this is a top level root module. Moreover, Data viewer should be imported from pyworkflow.em.viewer, since viewers_data.py is kind of the implementation file.

pyworkflow/em/__init__.py
```diff
@@ -77,16 +76,24 @@ def findSubClasses(classDict, className):
...
+    preferredClassViewer = Domain.getPreferredViewer(className)
+    preferredViewers = []
     for viewer in Domain.getViewers().values():
         if environment in viewer._environments:
             for t in viewer._targets:
                 if t in baseClasses:
-                    viewers.append(viewer)
+                    if viewer is preferredClassViewer:
+                        preferredViewers.insert(0, viewer)
+                    elif viewer is DataViewer:
```
delarosatrevin: Why DataViewer is treated as a special preferred viewer?

pyworkflow/gui/project/viewprotocols.py
```diff
@@ -485,15 +485,14 @@ def getObjectActions(self, obj):
 
         viewers = em.findViewers(obj.getClassName(), DESKTOP_TKINTER)
 
-        def yatevalejosemi(viewer):
+        def viewerCallback(viewer):
```
delarosatrevin: Hehehehe...it was time to give a proper name to this function.

pyworkflow/plugin.py
```diff
@@ -192,6 +192,28 @@ def getWizards(cls):
     def getName(cls):
         return cls._name
 
+    @classmethod
+    def loadPreferredViewers(cls):
+        if not hasattr(cls, '_preferredViewers'):
+            preferredViewers = os.environ.get('VIEWERS', {})
```
delarosatrevin: In one of the recently refactors, I tried to centralize the enviornment/config variables under pw.Config class. Would you mind to manage VIEWER this way?

pyworkflow/project/project.py
```diff
@@ -247,7 +247,7 @@ def load(self, dbPath=None, hostsConf=None, protocolsConf=None, chdir=True,
                 # we are loading a project after a Project.setDbName,
                 # used when running protocols
                 settingsPath = os.path.join(self.path, self.settingsPath)
-                print("settingsPath: ", settingsPath)
+                print("settingsPath: %s" % settingsPath)
```
delarosatrevin: Maybe we should just remove this debug print or only activate in SCIPION_DEBUG mode.

pyworkflow/em/__init__.py
delarosatrevin: I know that we have some other "dirty" import such as this one, but we should aim to remove them and make each component more modular, not add more mess.

yaizar: I didn't mean to create the PR so early but since I'll create it sooner or later I thought I'd better leave it here, I'm gonna push more changes, I'll assign reviewers when done

pyworkflow/em/__init__.py
yaizar: @pconesa thought it would be good that this one is the default if one isnt specified

pyworkflow/plugin.py
yaizar: yup, same as @pconesa said, wasnt aware of this one, iwll do

pyworkflow/project/project.py
yaizar: K

pyworkflow/em/__init__.py
delarosatrevin: If no one is specified, then the viewer will be the only one in the viewers list and will be picked up.

pyworkflow/plugin.py
```diff
@@ -192,6 +192,28 @@ def getWizards(cls):
...
+            if preferredViewers:
+                cls._preferredViewers = json.loads(preferredViewers)
+        return cls._preferredViewers
+
+    @classmethod
+    def getPreferredViewer(cls, className):
+        preferredViewerStr = cls.loadPreferredViewers().get(className, "")
+        if preferredViewerStr:
+            preferredViewerStr = preferredViewerStr.rsplit('.', 1)
+            preferredViewer = getattr(importlib.import_module(preferredViewerStr[0]),
```
delarosatrevin: Just one extra comment, why don't use the function importFromPlugin, that I think can handle the name even without splitting by '.' and also have some common way of reporting errors?

pyworkflow/plugin.py
```diff
@@ -192,6 +192,28 @@ def getWizards(cls):
...
+                                      preferredViewerStr[1],
+                                      None)
```
dmaluenda: If the `preferedViewerStr[0]` doesn't exist, this fails. Maybe a try/except statement instead of `if preferedVieweri is None`.

If the  `preferedViewerStr[1]` doesn't exist, this work fine (using the genral one from the Scipion).

delarosatrevin: try/except is for handling exception cases...I think is better the way it is...just checking if there is any defined string value.

dmaluenda: Yes, but this fails if a user puts a non valid module to be imported, or if the plugin is not installed.

The best option is just you said using importFromPlugin (but the spliting is still needed).
```
 preferredViewer = importFromPlugin(preferredViewerStr[0]), preferredViewerStr[1])
```

pyworkflow/plugin.py
yaizar: maybe your preferred viewer doesnt come from a plugin? first thing that comes to my mind, not sure if that function can treat pyworkflow as a plugin

pyworkflow/plugin.py
delarosatrevin: Yes, I agree that maybe it should be validated that the VIEWER variable contains valid viewers, but this should done once and not every time looking for viewers.

pyworkflow/em/__init__.py
yaizar: I don't think thats right, right now you get a list with all viewers targeting the class in whatever order the Domain.getViewers returns

pyworkflow/gui/project/viewprotocols.py
pconesa: don't worry, there are more stuff like that!

delarosatrevin:
Hi @yaizar, is this PR ready to be reviewed and merged?

yaizar:
Yes, it is now

delarosatrevin:
I'm not sure if I'm able to configure this properly for testing.

First, I exported an environment variable VIEWERS as follows:
```
export VIEWERS='{"Volume": ["bsoft.viewers.BsoftViewer"]}'
```
I checked it is recognized form python
```
>>> import pyworkflow as pw
>>> pw.Config.VIEWERS
{'Volume': ['bsoft.viewers.BsoftViewer']}
```

Then I run a protocol for crop/resize volumes (just a single volume output) and when clicking on Analize Results, I"m still getting our built-in DataViewer display. 

@yaizar , I'm missing something here?

yaizar:
I think you have the wrong path to the bsoft viewer, should be `["bsoft.viewers.bsoft_viewers.BsoftViewer"]`

delarosatrevin:
I don't think so @yaizar. 
Indeed, I think that we should use the import from package.viewers submodule where classes are exposed, not the internal files where Classes/Functions are implemented (e.g bsoft_viewer.py in this case). The same applies to 'protocols', 'objects', etc.

Anyway, I changed the path to check, but still getting the DataViewer as default one in "Analyze Results". Can you get someone else to test this in a different computer?

yaizar:
Can you double check you have bsoft installed? I can change the way we're making the comparison so that we try to import the preferred viewer (now it's doing string comparison)

yaizar:
AH SORRY SORRY SORRY. I didn't see you were doing it with analyze results. You'll get the Bsoft viewer if you double click on the volume. If you want bsoft viewer on analyze results you have to assign it to the protocol class @delarosatrevin

delarosatrevin:
Yes, I have Bsoft installed and the viewer works if I right-click and then explicitly select it. I think the comparison should not be done with strings. We should the class if it is the same. 

Regarding the "Analize Results": the default behaviour when no specific viewer is found for a protocol...it shows the view of the output. So, I think even if it is not defined for a protocol, to be consistent, the Analize Results should also show the preferred viewer of the output in this case.

yaizar:
@delarosatrevin @pconesa @dmaluenda this is ready for review, for me it works both double clicking and on Analyze Results if the protocol doesn't have its own assigned viewer. 
If user enters an invalid preferred viewer a message is printed in the console and the normal (i.e. old behavior) one will open.

delarosatrevin:
Still not working for me... :(
```
export VIEWERS='{"Volume": ["pyworkflow.em.viewers.DataViewer"]}'
```
But still BsoftViewer is the default one (both double-click and Analyze Results) for volume objects.

yaizar:
:(((( For me it works and I get exactly the opposite behavior (data viewer
if nothing specified or if data viewer specified, Bsoft viewer only if
bsoft specified). I asked @dmaluenda to take a look, I’ll try in other
machines too..

dmaluenda:
This is working fine for me. I check three option in both cases, by double-clicking a SetOfCoordinates and by "Analyze Results" button of XmippAutoPicking.

The three tested cases are:
1. The 'prefferedClassViewer' doesn't exist: a message is shown in terminal and a default viewer is used
2. The 'prefferedClassViewer' exists and it has defined as target what you pretend to show: Scipion uses that Viewer properly.
3. The 'prefferedClassViewer' existe but it has **NOT** defined as target what you pretend to show: nothing is print in terminal and the default viewer is used (or the first one in the list returned by findViewers())

Maybe in the last case, we can print a message in terminal. But, that's a minor.

All this have been done using the following lines in the `config/scipion.conf`
```
VIEWERS={"SetOfCoordinates": ["xmipp3.viewers.viewer.XmippViewer"],
         "XmippParticlePickingAutomatic": ["xmipp3.viewers.viewer.XmippViewer"]}
```

Also I think that we should allow to use not only list in the prefferedViewers but also single strings. I think that is more user friendly. Of course, I think that the use of list is a must to be able to describe more than one.

dmaluenda:
Another comment, it would be great if a protocol is a subclass of one in VIEWERS.keys(), we should use that viewer instead of the default one.

pconesa:
I've added the VARIABLE.....and sure I did something wrong....but scipion does not open. it shouldn't.

>>>>> python  /home/pablo/desarrollo/scipion/pyworkflow/apps/pw_manager.py 
  File "/home/pablo/desarrollo/scipion/pyworkflow/apps/pw_manager.py", line 32, in <module>
    from pyworkflow.gui.project import ProjectManagerWindow
  File "/home/pablo/desarrollo/scipion/pyworkflow/__init__.py", line 45, in <module>
    class Config:
  File "/home/pablo/desarrollo/scipion/pyworkflow/__init__.py", line 65, in Config
    VIEWERS = ast.literal_eval(__get('VIEWERS', "{}"))
  File "/home/pablo/desarrollo/scipion/software/lib/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/home/pablo/desarrollo/scipion/software/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    /home/pablo/desarrollo/scipion/{"Volume":["bsoft.viewers.BsoftViewer"],
    ^
SyntaxError: invalid syntax

pyworkflow/plugin.py
pconesa: We have tested and pyworkflow plugins are understood.

pyworkflow/em/__init__.py
pconesa: This is y only doubt here. If by chance (this has happened ot me before) the BSOFt volume viewer happens to be the first in the list.....it will be the default one.....unless the user write something in VIEWERS var?

pyworkflow/plugin.py
pconesa: I think this is addressed.

pyworkflow/em/__init__.py
pconesa: I don't see this happening here, so I'll merge this, but I still think that, unless the user defines something else......default viewer should DataViewer (just because the "subset creation capabilities")

PR #1918, opened 2019-02-07
Refactorization of outputs using _outputs List

pconesa:
Populate the outputList and store it.
iterOutputAttributes allows None as class to return all.
use iterOutputAttributes without EMObject and replace iterOutputsEM by
iterOutputAttributes

@delarosatrevin this is how I understood we should do this after our conversation.
I've unified calls to outputs iteration to a single one (iterOutputAttributes), but left iterOutputEM. Scipion is not using it, except in a test, but I guess plugins might be using them.

### DIFF
--- a/pyworkflow/protocol/protocol.py
+++ b/pyworkflow/protocol/protocol.py
@@ -25,11 +25,6 @@
 # **************************************************************************
 from __future__ import print_function
 
-import traceback
-
-from pyworkflow.em import EMObject
-from pyworkflow.utils import printTraceBack
-
 """
 This modules contains classes required for the workflow
 execution and tracking like: Step and Protocol
--- a/pyworkflow/protocol/protocol.py
+++ b/pyworkflow/protocol/protocol.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 
 import traceback
 
+from pyworkflow.em import EMObject
 from pyworkflow.utils import printTraceBack
 
 """
@@ -320,6 +321,9 @@ class Protocol(Step):
         self.mapper = kwargs.get('mapper', None)
         self._inputs = []
         self._outputs = CsvList()
+        # This flag will be used to annotate it output are already "migrated"
+        #  and available in the _outputs list. Therefore iterating
+        self._useOutputList = Boolean(False)
         # Expert level needs to be defined before parsing params
         self.expertLevel = Integer(kwargs.get('expertLevel', LEVEL_NORMAL))
         self._definition = Form(self)
@@ -415,7 +419,14 @@ class Protocol(Step):
                 self._deleteChild(k, v)
             self._insertChild(k, v)
 
+        # If outputs not migrated to output list
+        self._migrateOutputs()
+
+        # Store attributes in _output (this does not persist them!)
         self._storeAttributes(self._outputs, kwargs)
+
+        # Persist outputs list
+        self._insertChild("_outputs", self._outputs)
 
     def _updateOutputSet(self, outputName, outputSet,
                          state=Set.STREAM_OPEN):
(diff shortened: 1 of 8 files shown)


pyworkflow/protocol/protocol.py
```diff
@@ -27,6 +27,7 @@
 
 import traceback
 
+from pyworkflow.em import EMObject
```
delarosatrevin: I don't like this import here...pw.protocol is more general that pw.em

pyworkflow/protocol/protocol.py
```diff
@@ -635,22 +646,83 @@ def hasLinkedInputs(self):
...
         return (linkedPointers and not emptyPointers)
 
-    def iterOutputAttributes(self, outputClass):
+    def iterOutputAttributes(self, outputClass=None):
         """ Iterate over the outputs produced by this protocol. """
-        for key, attr in self.getAttributes():
-            if isinstance(attr, outputClass):
+
+        # If used with EMObject it's an obsolete code
+        if outputClass == EMObject:
```
delarosatrevin: I don't think this validation is useful, we can remove it and make sure it is not longer used with EMObject.

pyworkflow/protocol/protocol.py
```diff
@@ -635,22 +646,83 @@ def hasLinkedInputs(self):
...
+
+            # Get it from the protocol
+            attr = getattr(self, attrName)
+
+            yield attrName, attr
+
+    def _iterOutputsOld(self):
+        """ This method iterates assuming the old model: any EMObject attribute
+        is an output."""
+        from pyworkflow.em import EMObject
```
delarosatrevin: This import is duplicated with the top one, I don't like none, but probably this one inside is acceptable since we should deprecate this way at some point.

pyworkflow/protocol/protocol.py
```diff
@@ -635,22 +646,83 @@ def hasLinkedInputs(self):
...
+            outputClass = None
+
+        iterator = self._iterOutputsNew if self._useOutputList else self._iterOutputsOld
+
+        # Iterate
+        for key, attr in iterator():
+            if outputClass is None or attr is isinstance(attr, outputClass):
                 yield key, attr
 
     def iterOutputEM(self):
```
delarosatrevin: We can just delete this function and update the test that is currently using it.

pyworkflow/protocol/protocol.py
```diff
@@ -415,8 +419,15 @@ def _defineOutputs(self, **kwargs):
                 self._deleteChild(k, v)
             self._insertChild(k, v)
 
+        # If outputs not migrated to output list
+        self._migrateOutputs()
```
delarosatrevin: Some we discuss was to not modify the protocol under the hood (without explicit action of the developer/user). I don't think we should migrate outputs...since it will not longer be needed for new protocols. 

I think it is missing the setting of the flag and the storing of the outputs attributes when calling the defineOutputs, that is an explicit action when the developer of a protocol is storing the results.

pyworkflow/protocol/protocol.py
pconesa: From scipion is safe..from the plugins...not sure.

pyworkflow/protocol/protocol.py
pconesa: in a defineOutput there are changes going on....in an streaming scenario I might even have output/s the old way. So, when are we going to use the new way?

pconesa: the useOutputList deafults to False....even for new created protocols. You want to keep 2 ways alive?

pyworkflow/protocol/protocol.py
delarosatrevin: This is the moment removed it and clean up. I just checked in the old scipion code and I haven't found iterOutputsEM in any package code. I doubt that new plugins use it, but if so, should only be a few cases we could easily fix.

pyworkflow/protocol/protocol.py
delarosatrevin: I don't get your point. What do you mean that in defineOutput are changes going on? Usually the define output is used once..and the output is updated in streaming, the output does not change that much from the non-streaming case. The new way will be used from the moment the new code is updated and get in use.

pyworkflow/protocol/protocol.py
pconesa: it's used in xmipp just one place.

pyworkflow/protocol/protocol.py
pconesa: I see, we had it before in protocol.py, locally.

pyworkflow/protocol/protocol.py
delarosatrevin: For newly created protocols...as soon as the defineOutput is used...it will behave in the new way. The 2 ways should be accepted for backward compatibility for some time.

pyworkflow/protocol/protocol.py
pconesa: ok, I'm removing the global import, and after release we can clean this EMObject thing.

pyworkflow/protocol/protocol.py
pconesa: No, the point it to test that iterOutputEM is still working as before...but a quick search just detected this usage in eman, plugin....should we get rid of EMObject and iterOutputEM?

PR #1920, opened 2019-02-08
Link Scalar outputs

pconesa:


pyworkflow/protocol/protocol.py
```diff
@@ -588,9 +588,14 @@ def iterInputAttributes(self):
                     # PointerList, this is used in viewprotocols.py
                     # to group them inside the same tree element
                     yield key, item
-            elif attr.isPointer() and attr.hasValue():
+            if attr.isPointer() and attr.hasValue():
                 yield key, attr
 
+            # Consider here scalars with pointers inside
+            elif isinstance(attr, Scalar) and attr.hasPointer():
+                if attr.get() is not None:
```
delarosatrevin: I think attr.get() should give you the value of the Pointed object.get(),but this also cover if the pointed is None?
Maybe we should add att.getPointer() (line below) instead of attr._pointer

pconesa: To the Scalar object you mean?

delarosatrevin: Yep, the Scalar object.

PR #1922, opened 2019-02-12
Allowing choosing pointers to Scalars in the GUI

pconesa:
Also, validation errors at runtime do not print the trace.

### DIFF
--- a/pyworkflow/gui/form.py
+++ b/pyworkflow/gui/form.py
@@ -1151,7 +1151,7 @@ class ParamWidget:
             self._onlyLabel = True
         else:
 
-            if not param.allowPointers:
+            if not param.allowsPointers:
                 var = tk.StringVar()
 
                 if issubclass(t, params.FloatParam) or issubclass(t, params.IntParam):
@@ -2127,7 +2127,7 @@ class FormWindow(Window):
             isinstance(param, params.RelationParam)):
             widgetValue = protVar
         # For Scalar params that allowPointers
-        elif param.allowPointers:
+        elif param.allowsPointers:
             if protVar.hasPointer():
                 # Get the pointer
                 widgetValue = protVar.getPointer()
@@ -2341,7 +2341,7 @@ class FormWindow(Window):
                     param.copy(value)
                 # Special treatment for Scalars that allow pointers
                 # Combo widgets do not have .param!
-                elif hasattr(widget, "param") and widget.param.allowPointers:
+                elif hasattr(widget, "param") and widget.param.allowsPointers:
                     if isinstance(value, pwobj.Pointer):
                         # Copy the pointer, otherwise changes in the
                         # widget pointer will be reflected
@@ -2349,13 +2349,13 @@ class FormWindow(Window):
                         pointerCopy.copy(value)
                         param.setPointer(pointerCopy)
                     else:
-                        param.set(value)
                         param.setPointer(None)
+                        param.set(value)
 
                 elif isinstance(param, pwobj.Object):
                     param.set(value)
             except ValueError:
-                if len(widget.get()):
+                if len(value):
                     print(">>> ERROR: setting param for: ", paramName,
                           "value: '%s'" % value)
                 param.set(None)
--- a/pyworkflow/gui/form.py
+++ b/pyworkflow/gui/form.py
@@ -2326,10 +2326,10 @@ class FormWindow(Window):
 
                 # Special treatment for pointer params
                 if isinstance(param, pwobj.Pointer):
-                    param.copy(widget)
+                    param.copy(value)
                 # Special treatment for Scalars that allow pointers
                 # Combo widgets do not have .param!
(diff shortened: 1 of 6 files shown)


pyworkflow/gui/form.py
```diff
@@ -1855,6 +2114,13 @@ def getWidgetValue(self, protVar, param):
             isinstance(param, params.MultiPointerParam) or
             isinstance(param, params.RelationParam)):
             widgetValue = protVar
+        # For Scalar params that allowPointers
+        elif param.allowPointers:
+            if protVar.hasPointer():
+                # NOTE: to check with JM, we should avoid accessing _pointer
+                widgetValue = protVar._pointer
```
delarosatrevin: I added the fuction, so now you could use: protVar.getPointer()

dmaluenda:
I'm getting the following error when I chose and select a setOfSomething for any input of any protocol in the paramWidget (browser).

```
Exception in Tkinter callback
  File "/home/malu/scipionRELEASE/software/lib/python2.7/lib-tk/Tkinter.py", line 1541, in __call__
    return self.func(*args)
  File "/home/malu/scipionRELEASE/pyworkflow/gui/form.py", line 1374, in _onVarChanged
    self.callback(self.paramName)
  File "/home/malu/scipionRELEASE/pyworkflow/gui/form.py", line 2248, in _checkChanges
    self.setParamFromVar(paramName)
  File "/home/malu/scipionRELEASE/pyworkflow/gui/form.py", line 2329, in setParamFromVar
    param.copy(widget)
  File "/home/malu/scipionRELEASE/pyworkflow/object.py", line 431, in copy
    self._copy(other, copyDict, copyId, ignoreAttrs=ignoreAttrs)
  File "/home/malu/scipionRELEASE/pyworkflow/object.py", line 457, in _copy
    self._objId = other._objId
AttributeError: ParamWidget instance has no attribute '_objId'
```

When I checkout devel-pluginization back, this doesn't happens. When I return to  `pc-scalar-pointe` the error is gotten.

dmaluenda:
I've just realized that this error is not happening on the Scipion protocols, only in plugin protocols.

pyworkflow/gui/form.py
pconesa: thanks, I'll change this to use the getter.

pconesa:
ummm, I'll double check that.

pconesa:
Got it, I've introduced this error while renaming variables. grrr!

pconesa:
[x] Pointer params (allowsNull and not): tested on extract particles (ctf for nullable pointer)
[x] PointerList params (adding/removing to the selection): tested on the summary monitor
[x] Scalar values are still possible to enter by typing, even if the 'allowsPointer' is set to True.
[x] Scalar values are properly used, both if set directly or by pointer.

pyworkflow/gui/form.py
pconesa: Done too!

dmaluenda:
I've tested and it work pretty fine in the following cases (except for the laast one):
[x] `allowPointers=True` in combination with `allowsNull=True` in the same parameter and giving an object.
[x] `allowPointers=True` in combination with `allowsNull=True` in the same parameter and giving a value.
[x] `allowPointers=True` in combination with `allowsNull=True` in the same parameter and giving nothing.
[x]  `allowPointers=True`, giving an object and then running the wizard. The wizard takes the object value (as expected), the same for the direct value (former way).
[ ]  `allowPointers=True` in combination with `allowsNull=True` in the same parameter where I putted an object and then it is deleted (resulting an empty field). **The protocol takes the last object**...

For the last case (which is failing since we expect a None value), I think it would be great if the trash icon is shown. I've deleted the object by putting the cursor there and pressing the 'del' key from the keyboard. I think that in this way the value remains loaded in the protocol class and it's not updated.

Very great job!!!

dmaluenda:
Just a minor, but useful I guess, I think we can standardize the option name. We have the ''allow**s**Null" option, so I will suggest the "allow**s**Pointer~s~".

pconesa:
Thank you @dmaluenda  for the detailed report. I'll rename the parameter adding an "s".. and check the behaviour  in combination with allowsNull.

dmaluenda:
@pconesa also have a look on how to delete an object if, finally, the user want to pass nothing.

pconesa:
In theory...emptying the box should empty the params

pconesa:
I rather avoid the "bin" for this case.

dmaluenda:
In theory, but as I pointed out above, when you select an object using the lens and, after, that object is deleted; the protocol is still taking that deleted object.

 > [ ] allowPointers=True in combination with allowsNull=True in the same parameter where I putted an object and then it is deleted (resulting an empty field). The protocol takes the last object ...

pconesa:
I'll look at that and make it work.

pconesa:
All addressed now, thank you both! (@delarosatrevin and @dmaluenda )

PR #1926, opened 2019-02-13
Add queuestepexecutor

ldelcano:
I think it can be merged and further tested.


pyworkflow/protocol/launch.py
```diff
@@ -124,7 +124,7 @@ def _launchLocal(protocol, wait, stdin=None, stdout=None, stderr=None):
     hostConfig = protocol.getHostConfig()
     useQueue = protocol.useQueue()
     # Check if need to submit to queue    
-    if useQueue:        
+    if useQueue and (protocol.getSubmitDict()["QUEUE_FOR_JOBS"] == "N"):
```
delarosatrevin: What about avoiding duplicated code and hiding if using 'Y' or 'N' (Btw, maybe a bit more verbose could be better, e.g 'Yes'/'No' or 'True'/'False', since it is already used in other config files) and creating a function in protocol?

```python
class Protocol:
    ...
    def useQueueForSteps(self):
        """ Return True if the protocol will use the queue for submitting each computing step. """
        return self.useQueue() and self.getSubmitDict()["QUEUE_FOR_STEPS"].lower() == "true"
```

pyworkflow/protocol/protocol.py
```diff
@@ -1997,10 +2004,15 @@ def runProtocolMain(projectPath, protDbPath, protId):
                                      hostConfig=hostConfig)
             sys.exit(retcode)
 
-        elif protocol.numberOfThreads > 1:
-            executor = ThreadStepExecutor(hostConfig,
-                                          protocol.numberOfThreads.get()-1,
-                                          gpuList=protocol.getGpuList())
+        elif nThreads > 1:
+            if protocol.useQueue() and (protocol.getSubmitDict()["QUEUE_FOR_JOBS"] == "Y"):
```
delarosatrevin:
```python
if protocol.useQueueForSteps():
    ....
```

pyworkflow/protocol/protocol.py
```diff
@@ -1997,10 +2004,15 @@ def runProtocolMain(projectPath, protDbPath, protId):
...
+                executor = QueueStepExecutor(hostConfig, protocol.getSubmitDict(), nThreads-1, gpuList = protocol.getGpuList())
+            else:
+                executor = ThreadStepExecutor(hostConfig, nThreads-1, gpuList = protocol.getGpuList())
+
+    if executor is None and protocol.useQueue() and (protocol.getSubmitDict()["QUEUE_FOR_JOBS"] == "Y"):
```
delarosatrevin:
```python
if executor is None and protocol.useQueueForSteps():
    ....
```

pyworkflow/utils/properties.py
```diff
@@ -128,6 +128,8 @@ class Message:
     LABEL_THREADS = 'Threads'
     LABEL_MPI = 'MPI'
     LABEL_QUEUE = 'Use queue?'
+    LABEL_QUEUE_FOR_JOBS = 'Use queue for Jobs?'
```
delarosatrevin: What about 'Use queue for Steps?'

pyworkflow/tests/em/workflows/test_parallel_gpu_queue.py
```diff
@@ -325,3 +349,40 @@ def testGpuMPI(self):
                                                 "Rel.2D GPU MPI",
                                                 doGpu=True, MPI=4)
         self._checkAsserts(relionGpu14)
+
+class Test_Queue_Steps(TestQueueBase):
```
delarosatrevin: Class name not PEP8 compliant.

pyworkflow/protocol/launch.py
ldelcano: Hi @delarosatrevin, the way to select this kind of execution is on the QueueDialog form that appears when you select to Use queue. There is a new parameter on the Queue parameters called QUEUE_FOR_JOBS that by default is 'N'. It could be 'No' as you suggested but it has to be a string cause the way this is coded only allowed this type. @pconesa and I talked about all this and thought it would be better to hide this option on the form since it can confuse normal users.

pyworkflow/protocol/protocol.py
ldelcano: @delarosatrevin this was the original way I started, adding another field useQueueForSteps (originally a second checkbox but it could have been integrated on a single combobox, but as I explained above we decided to go for letting the form as it is and show the parameter on the QueueDialog.

delarosatrevin: Here my comment is just to hide the internal logic into a function from the Protocol class, it is not about how it is shown to the users.

pyworkflow/protocol/launch.py
delarosatrevin: Ok, I see...it is fine then. I agree that will be better to hide it...and then if it is not in the hosts.conf file it will not be shown. Thanks for the clarification!

pyworkflow/protocol/protocol.py
ldelcano: Ah! ok, I see, I will then create a function.

pyworkflow/utils/properties.py
ldelcano: Ok, although what we run it is not steps but jobs (only steps with runJob will be sent to the queue)

pyworkflow/tests/em/workflows/test_parallel_gpu_queue.py
ldelcano: mmmm, I just followed naming convention like other classes on that test. I will then change them all.

ldelcano: mmmm, I just followed naming convention like other classes on that test. I will then change them all.

delarosatrevin: I don't think so ;) Most of the classes for Test does not contains _ , some test functions do.

ldelcano: Anyway, I have changed them all cause almost all of them contained _ (Test_Queue_ALL to TestQueueALL and so forth).

delarosatrevin: Ok, thanks @ldelcano

PR #1929, opened 2019-02-18
Aj correlation measures

ajimoreno:
I don't know if the change made in xmippLib.py is really necessary or not.


pconesa:
@ajimoreno will test if this is needed in the ghost. I suspect it is not.

pconesa:
I'm closing this one...if turns out that this is needed, we can reopen it.

PR #1932, opened 2019-02-20
Fixing bug which was overwriting the range of the frames in the creat…

DStrelak:
…ed movies
See also https://github.com/I2PC/scipion-em-xmipp/pull/56


pconesa:
@delarosatrevin , this identified by a test not passing, having the FramedRange wrong. This also was caused since I "propagated" the FramesRange in the SoM (SetOfMovies).copyInfo.
The problem here is that the copyInfo was happening always and we think this should happen only when there the set is created:
`if os.path.exists(setFile) and os.path.getsize(setFile) > 0:`

Do you mind to have a look if we have missed anything?

PR #1939, opened 2019-02-27
Fixing inconsistency in emFolder

### LINKED ISSUE #1938: Inconsistency in emFolder definition.
**Describe the bug**
There is an inconsistency in how emFolder is defined: 

* When software is installed the definition is as follow:
https://github.com/I2PC/scipion/blob/b5ae9a0b2844d8cd065697b413369b6712e1e217/pyworkflow/install/funcs.py#L264-L265
and
https://github.com/I2PC/scipion/blob/b5ae9a0b2844d8cd065697b413369b6712e1e217/pyworkflow/install/funcs.py#L300-L301

* Whereas in the plugin's home definition is
https://github.com/I2PC/scipion/blob/b5ae9a0b2844d8cd065697b413369b6712e1e217/pyworkflow/plugin.py#L226-L231

**To Reproduce**
1. Edit 'SCIPION_SOFTWARE' (at `<scipion>/config/scipion.conf`) or 'EM_ROOT' (at `~/.config/scipion/scipion.conf`) in such a way to get  'SCIPION_SOFTWARE'/em different to  'EM_ROOT'
2. Install a binary, e.g. `./scipion installb locsacle`
3. Run a test regarding that binary, e.g. `./scipion test locscale.tests.test_protocol_locscale`
4. See that is failing due to the LOCSCALE_HOME='EM_ROOT'/locscale-0.1 is not found, but it is in 'SCIPION_SOFTWARE'/em/locscale-0.1 

**Expected behavior**
I think that 'SCIPION_SOFTWARE' is thought to home the software that Scipion need such as *bin/\**, *lib/\**, *include/\**... whereas 'EM_ROOT' is to home the plugin's binaries. So, I would change 
https://github.com/I2PC/scipion/blob/b5ae9a0b2844d8cd065697b413369b6712e1e217/pyworkflow/install/funcs.py#L300-L301
to
```
def getEmFolder(): 
    return os.environ.get('EM_ROOT', 'software/em')
```

**Desktop (please complete the following information):**
 - Version [e.g. 2.0]

dmaluenda:
closes  #1938

### DIFF
--- a/pyworkflow/install/funcs.py
+++ b/pyworkflow/install/funcs.py
@@ -298,7 +298,8 @@ class Environment:
 
     @staticmethod
     def getEmFolder():
-        return '%s/em' % Environment.getSoftware()
+        defaultValue = '%s/em' % Environment.getSoftware()
+        return os.environ.get('EM_ROOT', defaultValue)
 
     @staticmethod
     def getEm(name):
--- a/pyworkflow/plugin.py
+++ b/pyworkflow/plugin.py
@@ -36,6 +36,7 @@ from collections import OrderedDict
 from abc import ABCMeta, abstractmethod
 
 import pyworkflow.utils as pwutils
+from pyworkflow.install import Environment
 
 
 class Domain:
@@ -228,7 +229,7 @@ class Plugin:
         to the default value.
         """
         cls._defineVar(varName,
-                       os.path.join(os.environ['EM_ROOT'], defaultValue))
+                       os.path.join(Environment.getEm(defaultValue)))
 
     @classmethod
     @abstractmethod


dmaluenda:
That's not good enough....

If a plugin's home is defined in the config, it should be installed there but, now it's installed in EM_ROOT (or SCIPION_SOFTWARE/em before this PR)...

pyworkflow/install/funcs.py
```diff
@@ -298,7 +298,8 @@ def getTmpFolder():
 
     @staticmethod
     def getEmFolder():
-        return '%s/em' % Environment.getSoftware()
+        defaultValue = '%s/em' % Environment.getSoftware()
+        return os.environ.get('EM_ROOT', defaultValue)
```
pconesa: It looks safe. one thing, how safe is to remove getEmPackagesFolder ?

dmaluenda: It make sense... but we shout ensure that this function is not used anywhere.

delarosatrevin: Should we centralize the access to environment variables? Should we access them from the pyworkflow.Config class?

pconesa: I like that idea.

PR #1949, opened 2019-03-14
Avoid environment in inits

pconesa:
This will just avoid using the environment at import time. for 2 variables.


pconesa:
wait, this started from fixes, review after fixes

pconesa:
this is breaking python3 installation. I'm looking at it.

pyworkflow/tests/model/tests_logs.py
```diff
@@ -4,8 +4,9 @@
 import logging
 import unittest
 
+from pyworkflow import Config
 from pyworkflow.utils.utils import getLineInFile, isInFile
-from pyworkflow.utils.log import ScipionLogger, LOG_FILE
+from pyworkflow.utils.log import ScipionLogger, getLogConfiguration
```
delarosatrevin: @pconesa we should keep pyworkflow.utils as the submodule to import...all other files underneath should be implementation of functionality. Could you check if 'log' is imported in the init? Previous line to this one should also be changed.

pyworkflow/utils/log.py
```diff
@@ -26,54 +26,53 @@
...
 import logging
 import logging.config
-from pyworkflow.utils.path import makeFilePath
-
-
 
-# Get general log file path
-LOG_FILE = os.path.join(os.environ['SCIPION_LOGS'], 'scipion.log')
+from pyworkflow import Config
+from pyworkflow.utils.path import makeFilePath
```
delarosatrevin: Same here...'path' should not be directly used for import

pyworkflow/tests/model/tests_logs.py
pconesa: Do you mean to add this "from log import *" at the utils.__init__?

delarosatrevin: Yep. And then from outside utils module every import should be at the level of pyworkflow.utils

pyworkflow/utils/log.py
pconesa: Should we do then:
import pyworkflow.utils as pwutils

and then ...pwutils.whatevertoreachthecode?

delarosatrevin: Yes, I personally prefer this style...but this is how you want to make the import. The other comment is more what is hide/exposed by the module, i.e, never import something from an internal file of the 'submodule', unless it is a well-defined sub-submodule...I think we have improved that...but still a few places with this kind of imports remaining.

pconesa: ok, doing it thanks

pconesa:
Done.

pconesa:
thanks to you! @delarosatrevin for reviewing it.

PR #1950, opened 2019-03-14
replace wget cmd with curl

yaizar:
Change the download command to avoid the zero size files when the download fails

### DIFF
--- a/pyworkflow/install/funcs.py
+++ b/pyworkflow/install/funcs.py
@@ -248,7 +248,10 @@ class Environment:
         else:
             self._libSuffix = 'dylib'
 
-        self._downloadCmd = ('wget -nv -c -O %(tar)s.part %(url)s\n'
+
+        # change from wget to curl to avoid empty file when download fails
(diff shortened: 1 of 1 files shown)


pconesa:
I've been testing this, the idea was to be more robust to downloads errors. Turns out that what we already had is also robust..I need to do more test but this might not improve anything? Or I'm missing the scenario where this would improve it.

azazellochg:
I had few times when the file didn't exist on server ending up with a few kB tgz file (with 404 header I guess) and then it tried to continue the installation..

pconesa:
I've tried that but never manage to end up with "bytes" in the file.....In that case, I think even curl will fail...unless, this cannot happen with curl.

delarosatrevin:
If there is no clear advantage, I will keep it as it is. Indeed, we are downloading to a temp file and then moving. Maybe we could concatenate both commands (wget .. || mv ..) so in case of error the temporary wrong-size file does not end as the expected file.

pconesa:
I'm leaving this out.

PR #1951, opened 2019-03-14
fixes to release

yaizar:


pyworkflow/plugin.py
```diff
@@ -139,6 +130,45 @@ def getPlugin(cls, name):
                             "not found" % name)
         return m
 
+    @classmethod
+    def _refreshPlugin(cls, name):
```
delarosatrevin: If this method is not internal to Domain, it should not have the _ prefix. (It is used from the PluginManager code)

fonsecareyna82: Uffff, sorry, you're right. 
Maybe we could put this method in utils.py and use it in a general way as refreshModule.
What do you think?

delarosatrevin: I think it is already in the correct place.

PR #1958, opened 2019-03-26
Scipion demo fix

pconesa:


pconesa:
It also contains a warning in case there are naming collisions among plugins

delarosatrevin:
@pconesa I'm reviewing this PR.

Just a minor recommendation, why we don't put the demo in a test?
* Now 'scipion help' does show demo (but I think it is better not to have it as special case)
* My first attempt was to 'scipion test --grep demo' or 'scipion test --grep wizard' (so it will be searchable)
* Moreover, it depends on a dataset, so as a test, the required dataset can be downloaded on the fly (as with other tests)

This is a minor thing, if it is much work, we can leave it as it is.

pconesa:
I'd rather keep it more "handy", although you are right that as a test we will have the dataset downloaded.

The other weird thing as a test is that it will launch a GUI...I think it would be the first case, right?
I'm not strongly against any of the options.

PR #1961, opened 2019-03-26
Fi small things

cossorzano:


pyworkflow/em/convert/image_handler.py
```diff
@@ -409,6 +409,9 @@ def truncateMask(self, inputFile, outputFile, newDim=None):
         self.__runXmippProgram('xmipp_transform_threshold',
                                '-i %s --select above 1 --substitute '
                                'value 1' % outputFile)
+
+    def createEmptyImage(self,fnOut,Xdim=1,Ydim=1,Zdim=1,Ndim=1):
```
delarosatrevin: Two PEP8 violations in a single line: 
* missing space after comma
* variable names should not start with Capital letter

delarosatrevin:
I'm fine with the changes in ImageHandler. I haven't tested the ones in atomic_struct

pyworkflow/em/convert/image_handler.py
```diff
@@ -416,10 +416,11 @@ def truncateMask(self, inputFile, outputFile, newDim=None):
...
 
     @classmethod
-    def createEmptyImage(cls, fnOut, xDim=1, yDim=1, zDim=1, nDim=1):
-        xmippLib.createEmptyFile(fnOut, xDim, yDim, zDim, nDim,
-                                 xmippLib.DT_FLOAT)
-    
+    def createEmptyImage(cls, fnOut, xDim=1, yDim=1, zDim=1, nDim=1,
+                         dataType=None):
+        dt = dataType or cls.DT_FLOAT
+        xmippLib.createEmptyFile(fnOut, xDim, yDim, zDim, nDim, dataType)
```
pconesa: Thanks @delarosatrevin , shouldn't we use here dt.

And...I guess moving the default value "out" of the method signature..is to avoid xmippLib references there?

delarosatrevin: Yes, it should be dt here...yes, that's the reason of it, I tried cls.DT_FLOAT, but this does not work in the argument list...I think we should use the DT_X defined in the class and the from xmippLib (although now they are the same)

PR #1966, opened 2019-04-08
Use of standard way to define version of pyworkflow

delarosatrevin:
Specify the version in a more standard way. 
As proposed in: https://www.python.org/dev/peps/pep-0396/


pconesa:
Thanks @delarosatrevin , I fail to see the benefit of this unless we take it further and "join" __version__ with our VERSION_2_0 constant.

Still I think we need our CORE invention, that is different to any "release version".

delarosatrevin:
The benefit is not much right now, but I think it is always good to follow standards. It would be fine for me if we just keep the __version__ updated with the release version.

PR #1968, opened 2019-04-26
Including postprocess in 3D section of protocol.conf

dmaluenda:
Xmipp (localSharpening) and LocScale already have this section and it make sense to be in the tree after refine instead of at the end if not defined here.

Also I propose to include RelionPostProcess protocol in https://github.com/scipion-em/scipion-em-relion/pull/52 PR.

It will looks like:
image

### DIFF
--- a/pyworkflow/templates/protocols.template
+++ b/pyworkflow/templates/protocols.template
@@ -48,6 +48,7 @@ Protocols SPA = [
 		{"tag": "protocol_group", "text": "Refine", "openItem": "False", "children": [
 			{"tag": "section", "text": "more", "openItem": "False", "children": []}
 		]},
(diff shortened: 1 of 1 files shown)


pconesa:
Thanks @dmaluenda , to be honest each plugin can decide were to put each of its protocols. Adding this line here will not warranty anything. It will serve as a "hint".

dmaluenda:
mmm, I'm not totally agree. Without including it in the scipion's protocols.conf, the section is append at the end and in a different way (notice the bold font and the no 'C' in green). It's just a minor esthetic thing but I propose to include it. Also to give the hint as you mention.

_Without this PR_:
image

pconesa:
Theoretically there shoudn't be a difference. umm

PR #1978, opened 2019-05-20
fi getDimensions implementation

rmarabini:
one line fix:

    def getDimensions(self):
        self.getDim()

should be
    def getDimensions(self):
        return self.getDim()


rmarabini:
Ignore this pull request

PR #1998, opened 2019-06-25
add  progressbar capability to Scipion

rmarabini:
Prints a progressbar. There are four predefined formats: dot, deafult, full, objedid

Esamples
```
dot
           .............................................
full
           [======                                  ] 1/3 (  33%) 2 to go
default
           Progress: [                                        ]   0%
objectid
         [                                        ] 0/3 (  0%) (objectId=33)          

```
Tests; 

./scipion test pyworkflow.tests.test_utils.TessProgressBar.test_dot
./scipion test pyworkflow.tests.test_utils.TessProgressBar.test_default
./scipion test pyworkflow.tests.test_utils.TessProgressBar.test_full
./scipion test pyworkflow.tests.test_utils.TessProgressBar.test_objectid

Example of use:
```

 from time import sleep
 from pyworkflow.utils import Progressbar

 progress = ProgressBar(1000, fmt=ProgressBar.FULL)
 for x in xrange(progress.total):
    progress.current += 1
    print progress()
    sleep(0.1)
 print progress.done()
```

### DIFF
--- a/pyworkflow/tests/test_utils.py
+++ b/pyworkflow/tests/test_utils.py
@@ -4,17 +4,18 @@
 Created on Mar 25, 2014
 
 @author: airen
+@author: roberto.marabini
 '''
-import pyworkflow as pw
+
+from __future__ import print_function
 
 from subprocess import Popen
 import pyworkflow.utils as pwutils
 from pyworkflow.utils.process import killWithChilds
 from pyworkflow.tests import *
 from pyworkflow.utils import utils, prettyDict
+from pyworkflow.utils import ProgressBar
 
-
-    
 class TestBibtex(BaseTest):
     """ Some minor tests to the bibtexparser library. """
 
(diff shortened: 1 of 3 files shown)


delarosatrevin:
Thanks @rmarabini, I have started taking a look at the class. 
Despite being  simple, I don't like the fact that the API is not that well defined (e.g access to internal variables and so on). I couldn't stop myself to take a look what is out there for progressbar in Python:
For example, have you looked at this one: 
https://github.com/niltonvolpato/python-progressbar

It seems quite good (both for Python 2 and 3). And quite simple to even embed into our own code. If you still think is best to proceed with your own implementation...I would like to propose some API and other minor changes. Even if for current usage is good enough, maybe having a general solution (using other implementation) might pay off in the long run.

rmarabini:
Hi,

   I looked to that class but I was not able to make it work as I would
like.
This module prints the progressbar and in general this is what a user wants
but...
in our case that means that he bar appears in a terminal not in the scipion
GUI stdout tab.
In order to get this result we need to print the message with the bar from
scipion-em-relion.

You may remember that I asked a couple of questions regarding how the
stdout is redirected to scipion-stdout-tab.

In summary, I do not dislike the module but I do not know how to make it
work and show the result in scipion GUI.

delarosatrevin:
Thanks for checking @rmarabini . I will take a look at the sys.stdout issue, since it is more general than the progressbar. Independently of what option we choose for this. I will keep you posted.

delarosatrevin:
So @rmarabini, are you happy with the final form of the Progressbar class? Should we go ahead and merge this now?

rmarabini:
try:
    from pyworkflow.utils import ProgressBar
except:
    from localrec.progressbar import ProgressBar

rmarabini:
OK, I just finished the review and I am OK with me. 
Whoever can merge this to devel do it so. Thanks

I added two small modifications to jose miguel code:

1) print a new line in the method finish so next print is in a different line (there is flag to inactivate this option)
2) add NOBAR so progressbar does not print a bar. Since scipion use a font with variable width, the bar length is not as informative as it should

and of course test are updated.

PR #2002, opened 2019-06-27
scatterPlot functionality added

rsanchezgarc:
Scatter plot added to plotter

### DIFF
--- a/pyworkflow/em/viewers/plotter.py
+++ b/pyworkflow/em/viewers/plotter.py
@@ -79,7 +79,7 @@ class EmPlotter(Plotter):
         """ Create an histogram. """
         self.hist(yValues, nbins, facecolor=color, **kwargs)
 
-    def plotScatterPlot(self, xValues, yValues, color='blue', **kwargs):
+    def plotScatter(self, xValues, yValues, color='blue', **kwargs):
         """ Create an scatter plot. """
         self.scatterP(xValues, yValues, c=color, **kwargs)
   
--- a/pyworkflow/em/viewers/plotter.py
+++ b/pyworkflow/em/viewers/plotter.py
@@ -78,7 +78,11 @@ class EmPlotter(Plotter):
     def plotHist(self, yValues, nbins, color='blue', **kwargs):
         """ Create an histogram. """
         self.hist(yValues, nbins, facecolor=color, **kwargs)
-    
+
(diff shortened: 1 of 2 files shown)


pyworkflow/em/viewers/plotter.py
```diff
@@ -78,7 +78,11 @@ def plotAngularDistributionFromMd(self, mdFile, title, **kwargs):
     def plotHist(self, yValues, nbins, color='blue', **kwargs):
         """ Create an histogram. """
         self.hist(yValues, nbins, facecolor=color, **kwargs)
-    
+
+    def plotScatterPlot(self, xValues, yValues, color='blue', **kwargs):
```
delarosatrevin: Can we just name it 'plotScatter' ? 'plotScatterPlot' seems a bit redundant to me.

pyworkflow/gui/plotter.py
```diff
@@ -167,6 +167,7 @@ def createSubPlot(self, title, xlabel, ylabel, xpos=None, ypos=None,
         self.last_subplot = a
         self.plot = a.plot
         self.hist = a.hist
+        self.scatterP= a.scatter
```
delarosatrevin: Please follow PEP8, it should be spaces around = sign.

pyworkflow/em/viewers/plotter.py
rsanchezgarc: sure!

dmaluenda:
@rsanchezgarc  keep in mind that this code update can't be used in the plugin until new Scipion's release. Then, if some code in scipion-em-xmipp needs it, it can't be released until new Scipion's release.

delarosatrevin:
@dmaluenda Now Scipion releases will be mainly the core, so I don't expect them to be so sparse as in the past.

dmaluenda:
Right, that's true. Even though, we must take into account that new Scipion code will not be able until it's released.

Of course it seems so obvious, but it's a bit tricky. Because a user can update a plugin just by a single click in the plugin manager. However, the Scipion update is not as easy as that. Then, user can wait more to update the Scipion's core than a certain plugin. Therefore, we need to properly manage the ScipionCore-PluginVersion incompatibilities. I know that it's a plugin-side concern but, just to point it out.

PR #2003, opened 2019-06-28
Updated protocols.conf to include the new view of Model Building

mmmtnez:
Dear all,
here you are the new views of Scipion menu.


azazellochg:
I though we don't use this template for protocols anymore, only to define the sections. Plugins should provide the protocols in their own protocols.conf. Am I right, @pconesa ?

pconesa:
You are right @azazellochg but "Scipion protocols" should be defined here. I guess the Emx one has to go out, right?

PR #2005, opened 2019-07-05
Refactoring proposal for ProgressBar class, fixed stdout mystery

delarosatrevin:
Dear @rmarabini , as my review of PR #1998 ...I'm proposing some re-factoring to your ProgressBar class. I didn't want to directly commit my changes into your branch, so I have created this other PR to discuss my proposal. 

Regarding the refactoring...I had two main concerns:
* There was a lot of boiler plate code for flush, print, increment, etc...which relates somehow to the second point here
* There was no clear API, so the access was done directly to the attributes both for getting or setting then. I have introduced a  minimal set of functions...if you need access to other properties, you can add getX/setX methods...although my personal recommendation is to keep it simple and minimal.

Regarding the mystery of stdout redirection...the issue was in your code. There is a STRONG recommendation in Python to not use mutable objects (e.g { }, [ ], etc) as default values of function...because this can introduce very weird issues if you modify the default value. This was the case here...you passed as default sys.stdout, but this was evaluated before the Protocol redirection of sys.stdout took place. The solution is easy once you have found the problem...use None as default and then inside the constructor, use sys.stdout if no other output was passed as argument.

### DIFF
--- a/pyworkflow/tests/test_utils.py
+++ b/pyworkflow/tests/test_utils.py
@@ -106,7 +106,7 @@ class TestGetListFromRangeString(BaseTest):
             self.assertEqual(o, pwutils.getListFromRangeString(s2))
 
 
-class TessProgressBar(unittest.TestCase):
+class TestProgressBar(unittest.TestCase):
 
     def caller(self, total, step, fmt, resultGold):
         ti = time.time()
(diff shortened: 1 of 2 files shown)


rmarabini:
I aprove this pull request and I will check it in my own pull request (

rmarabini:
PLease,

  do not confirm this merge, I will do it myself after I double checked it

PR #2008, opened 2019-07-09
Distinguish between Dx and Dy symmetries

pconesa:
Added constant to describe Dy symmetries (used by Cryosparc)

### DIFF
--- a/pyworkflow/em/constants.py
+++ b/pyworkflow/em/constants.py
@@ -78,11 +78,10 @@ AX_Y = 1
 AX_Z = 2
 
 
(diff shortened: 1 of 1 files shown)


rmarabini:
Let me summarize the changes before merge:

1) fix the following error:

  SYM_TETRAHEDRAL = 2  # T 222
  SYM_TETRAHEDRAL_Z3 = 2  # T  
That is, two different symmetries has the same value a"2" assigned. The values shopuld be different for different symmetries

2) Dyhedric symmetry  (Dn) is splitted in DXn and DYn (second rotation about x or y axis)

PR #2012, opened 2019-07-14
add rct section

### LINKED ISSUE #2011: missing views?
pyworkflow/templates/protocols.conf has only SPA section, where did the rest go? I'm looking for RCT section. Xmipp protocols.conf doesn't have it either.

azazellochg:
closes #2011


pconesa:
Hi, @delarosatrevin , @azazellochg , ithanks for this, we moved the whole section to xmipp plugin but didn't push it. Now, considering that scipion provides the 2 import protocols, I like this more and xmipp should add its protocols according to this.

PR #2014, opened 2019-07-16
Join set do not loose ids of first set

pconesa:
This PR has 3 small changes:
1.- The join sets was losing the ids as soon as there were conflicts. Now, if allowed, it will keep the id of the first set. This will allow the usage of other protocols that us the Id for matching items, at least for the first set that is being merged.
2.- Fixes a "bug"  in the _idCount of the sets
3.- Closes the bd after loading the project info in the project list window.

### DIFF
--- a/pyworkflow/em/constants.py
+++ b/pyworkflow/em/constants.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # **************************************************************************
 # *
 # * Authors:     J.M. De la Rosa Trevin (jmdelarosa@cnb.csic.es)
--- a/pyworkflow/em/protocol/protocol_sets.py
+++ b/pyworkflow/em/protocol/protocol_sets.py
@@ -142,8 +142,7 @@ class ProtUnionSet(ProtSets):
 
         # Renumber from the beginning if either the renumber option is selected
         # or we find duplicated ids in the sets
-        cleanIds = ((not self.ignoreDuplicates.get() and self.duplicatedIds())
-                        or self.renumber.get())
+        cleanIds = not self.ignoreDuplicates.get() and self.duplicatedIds()
 
         #TODO ROB remove ignoreExtraAttributes condition
         #or implement it. But this will be for Scipion 1.2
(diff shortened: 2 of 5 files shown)


pyworkflow/em/protocol/protocol_sets.py
```diff
@@ -171,13 +172,15 @@ def createOutputStep(self):
...
+                    if not cleanIds or setNum == 1:
                         newObj.setObjId(objId)
                 else:
                     newObj = obj
 
-                if cleanIds:
+                if (cleanIds and setNum > 1) or self.renumber.get():
                     newObj.cleanObjId()
+                print ("FINAL ID: %s" % newObj.getObjId())
+                print ("OBJ ID: %s" % objId)
```
dmaluenda: Please, remove the prints or put them under debug mode

pyworkflow/em/constants.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
dmaluenda: I think is not needed

pconesa: It is! there is a degree symbol in line 99

pyworkflow/em/protocol/protocol_sets.py
pconesa: upss

pyworkflow/em/constants.py
dmaluenda: Ok! You right. However, as it's in a comment, I would change
```
# in the YZ plane such that rotation about the X axis by ~110° 
```
for 
```
# in the YZ plane such that rotation about the X axis by ~110deg.
```

dmaluenda:
On the other hand, could I use this PR to introduce a bug fix when trying to join two SetOfCoordinates? The problem is _createSetOfCoordinates has a differnet signature than the _createSetOfAnything and it's fixed using the same that we did for the protocol_SubSet:
```
    # Overwrite SetOfCoordinates creation
    def _createSetOfCoordinates(self, suffix=''):
        coordSet = self.inputSets[0].get()
        micSet = coordSet.getMicrographs()
        return ProtSets._createSetOfCoordinates(self, micSet, suffix)
```

PR #2026, opened 2019-07-22
Minor changes introduced for Tomo

delarosatrevin:
### Changes:
* Introduced EXPOSE_ITEMS as static property in Set classes. It will be used in the GUI to allow (or not) to choose the subitems in the browser tree. Due to performance issues, it was (hardcoded) allowed only for SetOfVolumes and atomic structures. 
* Optional extra argument filterClassFunc to avoid some classes when operating with sets.


pconesa:
Can this branch be deleted? @delarosatrevin

delarosatrevin:
Yep, done!

PR #2032, opened 2019-08-07
Json multipointer to sets

pconesa:
json export test
Validate PointerList

### DIFF
--- a/pyworkflow/protocol/protocol.py
+++ b/pyworkflow/protocol/protocol.py
@@ -1799,12 +1799,8 @@ class Protocol(Step):
                         paramErrors.append('cannot be EMPTY.')
                     # Consider empty pointers
                     else:
-                        for pointer in attr:
-                            obj = pointer.get()
-                            if obj is None:
-                                paramErrors.append('cannot have EMPTY items.')
-                                break
-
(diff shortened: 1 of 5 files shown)


pyworkflow/protocol/protocol.py
```diff
@@ -1792,6 +1792,20 @@ def validate(self):
...
                 if condition and obj is None and not param.allowsNull:
                     paramErrors.append('cannot be EMPTY.')
+            elif isinstance(attr, PointerList):
+                # In this case allowsNull refers to not allowing empty items
+                if not param.allowsNull:
+                    if len(attr) == 0:
+                        paramErrors.append('cannot be EMPTY.')
+                    # Consider empty pointers
+                    else:
+                        for pointer in attr:
```
delarosatrevin: Maybe this can be re-written in a more pythonic way:
```python
if any(pointer.get() is None for pointer in attr):
    paramErrors.append('Can not have EMPTY items.')
```

Note: I have not tested the above code.

pconesa: Thanks @delarosatrevin , I'll change it and test it.

PR #2042, opened 2019-08-24
Textstring was behaving like string

cossorzano:


pconesa:
HI @cossorzano , could you please explain which case are you trying to cover or what triggered this.
I thought Text params have been working fine so far.

cossorzano:
StringParam and TextParam both had the same appearance in the form (a single line textbox). However, TextParam was originally intended to hold "Text" (newlines included). This is particularly useful for long entries in which it makes sense to split them in several lines.

pconesa:
Thanks @cossorzano , I'll review this now I get it.

PR #2044, opened 2019-08-26
* Stop a workflow from a selected protocol

fonsecareyna82:


pconesa:
This PR should have gone to devel instead. 

Let's:
1.- merge first devel-fixes -->devel
2.- And then this one + PR devel-fixes -->devel (this time should have only this changes)

PR #2077, opened 2020-04-06
Ds fixing movie shift plot

DStrelak:
This PR lifts restriction we posed on graph generation, mainly equal axis scale.
As a result, it should automatically scale the axis, and making plots with dominant direction more easy to read, as in example bellow (left before, right after [note: it plots two different data sets, so the ranges differ]):
image

### DIFF
--- a/pyworkflow/em/protocol/protocol_align_movies.py
+++ b/pyworkflow/em/protocol/protocol_align_movies.py
@@ -689,26 +689,13 @@ def createAlignmentPlot(meanX, meanY):
 
     ax = figure.add_subplot(111)
     ax.grid()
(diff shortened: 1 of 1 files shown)


pconesa:
@DStrelak , I guess you are not yet using scipion3 right. this scipion repo should be frozen and all work done at:
https://github.com/scipion-em/scipion-pyworkflow
https://github.com/scipion-em/scipion-em
https://github.com/scipion-em/scipion-app

I can take your changes there if you want.

pconesa:
By the way, much better with your change!

DStrelak:
> I can take your changes there if you want.

That would be much appreciated!

